PR #1, opened 2020-04-18
Add wgpu_set_log_callback

almarklein:
This is a proposal to make it possible for projects that wrap the wgpu-native static library, to get the log messages via a callback.

Having them printed to stdout is much easier, but having a callback gives the application much more control (e.g. feeding log messages in the environments own logging system).

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -703,6 +703,8 @@ typedef uint32_t WGPUBackendBit;
 
 typedef void (*WGPURequestAdapterCallback)(WGPUOption_AdapterId id, void *userdata);
 
+typedef void (*WGPULogCallback)(const char *msg);
+
 typedef struct {
   WGPUOption_TextureViewId view_id;
 } WGPUSwapChainOutput;
@@ -989,6 +991,8 @@ void wgpu_request_adapter_async(const WGPURequestAdapterOptions *desc,
                                 void *userdata);
 
 void wgpu_sampler_destroy(WGPUSamplerId sampler_id);
+
+void wgpu_set_log_callback(int level, WGPULogCallback callback);
 
 WGPUSwapChainOutput wgpu_swap_chain_get_next_texture(WGPUSwapChainId swap_chain_id);
 
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -703,7 +703,7 @@ typedef uint32_t WGPUBackendBit;
 
 typedef void (*WGPURequestAdapterCallback)(WGPUOption_AdapterId id, void *userdata);
 
-typedef void (*WGPULogCallback)(const char *msg);
+typedef void (*WGPULogCallback)(int level, const char *msg);
 
 typedef struct {
   WGPUOption_TextureViewId view_id;
@@ -992,7 +992,9 @@ void wgpu_request_adapter_async(const WGPURequestAdapterOptions *desc,
 
 void wgpu_sampler_destroy(WGPUSamplerId sampler_id);
 
-void wgpu_set_log_callback(int level, WGPULogCallback callback);
+void wgpu_set_log_callback(WGPULogCallback callback);
+
+void wgpu_set_log_level(int level);
 
 WGPUSwapChainOutput wgpu_swap_chain_get_next_texture(WGPUSwapChainId swap_chain_id);
 
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -127,6 +127,15 @@ typedef enum {
 } WGPULoadOp;
 
 typedef enum {
+  WGPULogLevel_Off = 0,
+  WGPULogLevel_Error = 1,
+  WGPULogLevel_Warn = 2,
+  WGPULogLevel_Info = 3,
+  WGPULogLevel_Debug = 4,
+  WGPULogLevel_Trace = 5,
+} WGPULogLevel;
+
+typedef enum {
   WGPUPowerPreference_Default = 0,
   WGPUPowerPreference_LowPower = 1,
   WGPUPowerPreference_HighPerformance = 2,
@@ -994,7 +1003,7 @@ void wgpu_sampler_destroy(WGPUSamplerId sampler_id);
 
 void wgpu_set_log_callback(WGPULogCallback callback);
 
-void wgpu_set_log_level(int level);
+int wgpu_set_log_level(WGPULogLevel level);
 
 WGPUSwapChainOutput wgpu_swap_chain_get_next_texture(WGPUSwapChainId swap_chain_id);
 
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use log::{Record, Metadata};
+use std::ffi::CString;
+use std::os::raw::c_char;
+use std::os::raw::c_int;
+
+pub type LogCallback = unsafe extern "C" fn(msg: *const c_char);
+
+struct LogProxy {level: log::Level, callback: LogCallback}
+
+impl log::Log for LogProxy {
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        metadata.level() <= self.level
+    }
+
+    fn log(&self, record: &Record) {
+        if self.enabled(record.metadata()) {
+            let msg = format!("{} - {}", record.level(), record.args());
+            // println!("{}", msg);  // Print the message to stdout
+            let cb: LogCallback = self.callback;
+            let s = CString::new(msg).unwrap();
+            let p = s.as_ptr();
+            unsafe {
+                cb(p);
+            }
+            // We do not use std::mem::forget(s), so the memory will be reclaimed
+            // by Rust once this function here. The callback should thus make a
+            // copy of the string. This is fine, because its probably converted to
+            // an internal (unicode) string type.
+        }
+    }
+
+    fn flush(&self) {}
+}
+
+// Store the logger instance as a mutable nullable static object
+static mut LOGGER: Option<LogProxy> = None;
+
+#[no_mangle]
+pub unsafe extern "C" fn wgpu_set_log_callback(level: c_int, callback: LogCallback) {
+    // Get level
+    let level_filter = match level {
+        0 => log::LevelFilter::Off,
+        1 => log::LevelFilter::Error,
+        2 => log::LevelFilter::Warn,
+        3 => log::LevelFilter::Info,
+        4 => log::LevelFilter::Debug,
+        _ => log::LevelFilter::Trace,
+    };
+    // Instantiate logger and store as the static object
+    let logger: LogProxy = LogProxy{level: level_filter.to_level().unwrap(), callback: callback};
+    LOGGER = Some(logger);
+    // set_logger seems to require that the logger is static,
+    // so we need to use LOGGER instead of logger.
+    match &LOGGER {
+        Some(logger) =>log::set_logger(logger)
+            .map(|()| log::set_max_level(level_filter)).unwrap(),
+        None => println!("{}", "Failed to set logger"),
+    }
+}
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -1,7 +1,3 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
 use log::{LevelFilter, Metadata, Record};
 use std::ffi::CString;
 use std::os::raw::c_char;
@@ -37,11 +33,28 @@ impl log::Log for LogProxy {
 // Store the logger instance as a mutable nullable static object
 static mut LOGGER: Option<LogProxy> = None;
 
+#[repr(C)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum LogLevel {
+    Off = 0,
+    Error = 1,
+    Warn = 2,
+    Info = 3,
+    Debug = 4,
+    Trace = 5,
+}
+
 // Note: this function may only be called once in the lifetime of a program.
 #[no_mangle]
 pub unsafe extern "C" fn wgpu_set_log_callback(callback: LogCallback) {
+    // Check if the logger has already been set
+    match &LOGGER {
+        Some(_) => panic!("The logger callback can only be set once."),
+        None => (),
+    }
     // Instantiate logger, store as static object, and make it THE logger.
-    let logger: LogProxy = LogProxy { callback: callback };
+    let logger: LogProxy = LogProxy { callback };
     LOGGER = Some(logger);
     log::set_logger(LOGGER.as_ref().unwrap()).unwrap();
     // The max_level is Off by default. If not set yet, set it to Warn instead.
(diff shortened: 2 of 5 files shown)


src/logging.rs
```diff
@@ -0,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
almarklein, line 1: I put this logic in a new file because it did not seem to fit in anywhere else. I'd be happy to move it elsewhere tho.

src/logging.rs
```diff
@@ -0,0 +1,63 @@
...
+struct LogProxy {level: log::Level, callback: LogCallback}
+
+impl log::Log for LogProxy {
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        metadata.level() <= self.level
+    }
+
+    fn log(&self, record: &Record) {
+        if self.enabled(record.metadata()) {
+            let msg = format!("{} - {}", record.level(), record.args());
```
almarklein, line 21: I made the format such that the receiving end can decide what to do. But now that I read this back, we could instead give the callback two args `(level, msg)`.

src/logging.rs
```diff
@@ -0,0 +1,63 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use log::{Record, Metadata};
+use std::ffi::CString;
+use std::os::raw::c_char;
+use std::os::raw::c_int;
+
+pub type LogCallback = unsafe extern "C" fn(msg: *const c_char);
+
+struct LogProxy {level: log::Level, callback: LogCallback}
```
kvark, line 12: nit: can we put each field on a separate line?

src/logging.rs
kvark: yes, please, let's return 2 arguments back

src/logging.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        metadata.level() <= self.level
+    }
+
+    fn log(&self, record: &Record) {
+        if self.enabled(record.metadata()) {
+            let msg = format!("{} - {}", record.level(), record.args());
+            // println!("{}", msg);  // Print the message to stdout
+            let cb: LogCallback = self.callback;
+            let s = CString::new(msg).unwrap();
```
kvark, line 24: even better: we could return the length of the string explicitly as well as the string (aka Pascal format), so that no conversion is needed here (it's pretty expensive to heap-allocate a new CString).

src/logging.rs
```diff
@@ -0,0 +1,63 @@
...
+    };
+    // Instantiate logger and store as the static object
+    let logger: LogProxy = LogProxy{level: level_filter.to_level().unwrap(), callback: callback};
+    LOGGER = Some(logger);
+    // set_logger seems to require that the logger is static,
+    // so we need to use LOGGER instead of logger.
+    match &LOGGER {
+        Some(logger) =>log::set_logger(logger)
+            .map(|()| log::set_max_level(level_filter)).unwrap(),
+        None => println!("{}", "Failed to set logger"),
```
kvark, line 61: how could this be?

src/logging.rs
```diff
@@ -0,0 +1,63 @@
...
+        3 => log::LevelFilter::Info,
+        4 => log::LevelFilter::Debug,
+        _ => log::LevelFilter::Trace,
+    };
+    // Instantiate logger and store as the static object
+    let logger: LogProxy = LogProxy{level: level_filter.to_level().unwrap(), callback: callback};
+    LOGGER = Some(logger);
+    // set_logger seems to require that the logger is static,
+    // so we need to use LOGGER instead of logger.
+    match &LOGGER {
```
kvark, line 58: You probably want something like `log::set_logger(LOGGER.as_ref().unwrap())`

src/logging.rs
```diff
@@ -0,0 +1,63 @@
...
+    let level_filter = match level {
+        0 => log::LevelFilter::Off,
+        1 => log::LevelFilter::Error,
+        2 => log::LevelFilter::Warn,
+        3 => log::LevelFilter::Info,
+        4 => log::LevelFilter::Debug,
+        _ => log::LevelFilter::Trace,
+    };
+    // Instantiate logger and store as the static object
+    let logger: LogProxy = LogProxy{level: level_filter.to_level().unwrap(), callback: callback};
```
kvark, line 54: multi-line please! (did you try running `cargo fmt` on this?)

ffi/wgpu.h
```diff
@@ -994,6 +996,8 @@ void wgpu_request_adapter_async(const WGPURequestAdapterOptions *desc,
 
 void wgpu_sampler_destroy(WGPUSamplerId sampler_id);
 
+void wgpu_set_log_callback(int level, WGPULogCallback callback);
```
windelbouwman, line 999: Why not register a function which gets called with each log message in the form of a level/message pair?

ffi/wgpu.h
```diff
@@ -707,6 +707,8 @@ typedef uint32_t WGPUBackendBit;
 
 typedef void (*WGPURequestAdapterCallback)(WGPUOption_AdapterId id, void *userdata);
 
+typedef void (*WGPULogCallback)(const char *msg);
```
windelbouwman, line 710: I would expect 
`typedef void (*WGPULogCallback)(LogLevel level, const char *msg)`

src/logging.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    fn flush(&self) {}
+}
+
+// Store the logger instance as a mutable nullable static object
+static mut LOGGER: Option<LogProxy> = None;
+
+#[no_mangle]
+pub unsafe extern "C" fn wgpu_set_log_callback(level: c_int, callback: LogCallback) {
```
windelbouwman, line 43: I would expect two API functions:
- `wgpu_set_log_callback(callback: LogCallback)`
- `wgpu_set_log_level(level: c_int)`

ffi/wgpu.h
almarklein: Yes, this would indeed be better :)  though you also want to set the max log level to avoid the overhead of debug messages being sent over.

src/logging.rs
almarklein: Makes sense. Though combining them in a single function also works IMO. What do you think @kvark ?

ffi/wgpu.h
windelbouwman: Yes, true, so the level should be set at reasonable level by default (info or warning).

src/logging.rs
almarklein: The API uses `WGPURawString` (`const char *`) in other places where strings are communicated. Using a different format here would make the API less consistent. Do yo think it's worth it? (Your call. I just want to make sure we're on the same page :) )

src/logging.rs
kvark: I'm trying to imagine why anyone would change the log level at run-time and I can't. @windelbouwman do you have a use case in mind that would require them to be separate functions?
It seems to me that a single one is good.

almarklein: I can imagine it's use in an interactive environment (like Python or Julia), where something does not work, you set the log level via the repl, and then try again to see what's going on.

almarklein: Also unit testing: Some tests could crank up the log level, do some stuff while capturing logs, and inspect the logs to see if things are looking good. Frankly, now that I've implemented it, and had a feel of it working, I prefer to have the two functions :)

almarklein:
> did you try running `cargo fmt` on this?

Thanks for the tip; hooray for autoformatting :)

> You probably want something like log::set_logger(LOGGER.as_ref().unwrap())

Ah yes, that's much simpler.

> let's return 2 arguments back

Done. The callback now receives both the level and the message.

src/logging.rs
windelbouwman: I have no use case, it's just that I think setting the callback and setting the level are two different things, and therefore I would expect two seperate API functions.

src/logging.rs
kvark: The problem is that heap-allocation here is too heavy if an application logs a log and still wants to run smoothly at run-time. I guess we can have that now and then improve without breaking the API. For example, we could have a thread-local `Vec<u8>` that we put the string contents in and append a null terminator before calling the user function. Again, we can follow up with that, it's fine to proceed with what you have now.

src/logging.rs
kvark: two functions it is then!

src/logging.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl log::Log for LogProxy {
+    fn enabled(&self, _metadata: &Metadata) -> bool {
+        true
+    }
+
+    fn log(&self, record: &Record) {
+        if self.enabled(record.metadata()) {
+            let callback: LogCallback = self.callback;
+            let msg = format!("{}", record.args());
```
windelbouwman, line 24: Could this be replaced by `let msg = record.args().to_string();`?

almarklein:
> The problem is that heap-allocation here is too heavy if an application logs a log and still wants to run smoothly at run-time. I guess we can have that now and then improve without breaking the API. For example, we could have a thread-local Vec<u8> that we put the string contents in and append a null terminator before calling the user function. Again, we can follow up with that, it's fine to proceed with what you have now.

Cool, that sounds great! Even though I'm not sure if I follow :) Isn't this what `CString` basically does?

> Another thing that we may be missing is `wgpu_set_log_filter(source_prefix)`. [...] Is this something you'd want to introduce here, or follow-up?

Let's add it now while we're in here.

kvark:
> Isn't this what CString basically does?

CString makes a new allocation. What I'm describing is a way to retain a persistent allocation per thread. The thread locality allows you to modify it (i.e. fill it with the string and append null). So the cost becomes just a copy of the string, plus accessing the thread storage, but no allocation.

> Let's add it now while we're in here.

Worth noting that you may want multiple prefixes to be enabled. Not sure what the best API for that will be.

almarklein:
> Me: Let's add it now while we're in here.

I'm getting back at this. I had a try, but things are getting trickier, and I currently lack the Rust skills to make this work in a reasonable amount of time, sry!  So if we can go along with what we have now, that'd be great :)

kvark:
Alright, could you squash some of the commits to have the history be more linear and straightforward?

almarklein:
squashed!

src/logging.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Store the logger instance as a mutable nullable static object
+static mut LOGGER: Option<LogProxy> = None;
+
+// Note: this function may only be called once in the lifetime of a program.
+#[no_mangle]
+pub unsafe extern "C" fn wgpu_set_log_callback(callback: LogCallback) {
+    // Instantiate logger, store as static object, and make it THE logger.
+    let logger: LogProxy = LogProxy { callback: callback };
```
windelbouwman, line 44: If you run the clippy tool on this file, it would suggest using:
```rust
let logger: LogProxy = LogProxy { callback };
```

ffi/wgpu.h
```diff
@@ -994,6 +996,10 @@ void wgpu_request_adapter_async(const WGPURequestAdapterOptions *desc,
 
 void wgpu_sampler_destroy(WGPUSamplerId sampler_id);
 
+void wgpu_set_log_callback(WGPULogCallback callback);
+
+void wgpu_set_log_level(int level);
```
windelbouwman, line 1001: Consider changing `int` to an enum type here like this:
```c
typedef enum {
  WGPULogLevel_DEBUG = 0,
  WGPULogLevel_INFO = 1,
  // etc.
} WGPULogLevel

void wgpu_set_log_level(WGPULogLevel level);
```

src/logging.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn wgpu_set_log_level(level: c_int) {
+    let level_filter = match level {
+        0 => LevelFilter::Off,
+        1 => LevelFilter::Error,
+        2 => LevelFilter::Warn,
+        3 => LevelFilter::Info,
+        4 => LevelFilter::Debug,
+        _ => LevelFilter::Trace,
```
windelbouwman, line 61: Assume trace, or return an error here?

src/logging.rs
```diff
@@ -0,0 +1,64 @@
...
+
+// Store the logger instance as a mutable nullable static object
+static mut LOGGER: Option<LogProxy> = None;
+
+// Note: this function may only be called once in the lifetime of a program.
+#[no_mangle]
+pub unsafe extern "C" fn wgpu_set_log_callback(callback: LogCallback) {
+    // Instantiate logger, store as static object, and make it THE logger.
+    let logger: LogProxy = LogProxy { callback: callback };
+    LOGGER = Some(logger);
```
windelbouwman, line 45: The mutation of `LOGGER` happens in an unsafe function, consider creating another function, which is safe, which contains all code that must be safe, such as changing this global variable.

src/logging.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    fn flush(&self) {}
+}
+
+// Store the logger instance as a mutable nullable static object
+static mut LOGGER: Option<LogProxy> = None;
+
+// Note: this function may only be called once in the lifetime of a program.
```
windelbouwman, line 40: Instead of this comment here, consider returning an error code when setting the callback, such that the calling code can check whether this succeeded or not.

almarklein: We cannot really return an error in a C function. Or do you mean returning -1 or something?

ffi/wgpu.h
almarklein: Makes sense, I think. Would it make sense to copy over the `log::Level` enum, or perhaps a subset?

src/logging.rs
almarklein: If we do the enum in the C api, then we should be strict here, IMO.

src/logging.rs
almarklein: The `log::set_logger` call would panic, I suppose, when called twice, and I imagine that the error messages is quite similar to what we'd do.

kvark:
bors r-
looks like there are more changes to do

src/logging.rs
almarklein: Ok, maybe not as descriptive: `thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: SetLoggerError(())', src\libcore\result.rs:1165:5`

ffi/wgpu.h
kvark: copying the enum seems fine to me, if it's not already repr(C)

src/logging.rs
windelbouwman: Exactly, returning a value to C might be a good idea, at least something like `WGPU_OK` or `WGPU_ERROR` if this exists already.

src/logging.rs
almarklein: > If you run the clippy tool ...

```
λ rustup component add clippy
error: component 'clippy' for target 'x86_64-pc-windows-msvc' is unavailable for download
```
:(

src/logging.rs
almarklein: I tried, but I get `use of mutable static` for `LOGGER`, and I have no idea how to deal with that.

src/logging.rs
almarklein: Mmm, I applied this to `wgpu_set_log_level`, but for `wgpu_set_log_callback` I think I prefer to panic instead. Better fail hard (with a good error message) than fail silently because the caller fails to check the return value.

src/logging.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum LogLevel {
+    Off = 0,
+    Error = 1,
+    Warn = 2,
+    Info = 3,
+    Debug = 4,
+    Trace = 5,
+}
```
almarklein, line 46: @kvark is the the way to go? (I copied this from `wgpu-types`)

src/logging.rs
windelbouwman: Exactly, so the best way to fail in python is to check the result code, and raise an exception ;)

src/logging.rs
kvark: Yes, this look good!

kvark:
Are we all set? Or is there another change pending?

src/logging.rs
almarklein: Yeah but there is a risk of forgetting to check it. Also because this pattern is not common in the rest of wgpu. When an application calls this function more than once, it does something really wrong, and it deserves a panic ;)

almarklein:
Ready from my end, Windel?

almarklein:
ready!

PR #3, opened 2020-04-20
Remove license from source headers

### LINKED ISSUE #2: Remove source headers
```rust
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```

This block was needed for vendoring in mozilla-central. As the code is now moved out from `wgpu` to `wgpu-native`, it's no longer vendored, so we can remove the nasty headers.

tmpvar:
> This block was needed for vendoring in mozilla-central. As the code is
> now moved out from wgpu to wgpu-native, it's no longer vendored, so we
> can remove the nasty headers.

fixes #2

Let me know if I missed something!


kvark:
Bah, we basically don't have any CI coverage here right now :( Created #4

PR #13, opened 2020-04-23
Add wgpu_get_version()

almarklein:
This ads a C api function to obtain the version of `gpu-native`.

The use-case for wgpu-py is that although we do ship a pre-compiled wgpu-native library, we allow our users to bring their own library. We would like to check whether the version that the wrapper expects matches the actual library version :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,3 +13,12 @@ type Global = core::hub::Global<core::hub::IdentityManagerFactory>;
 lazy_static::lazy_static! {
     static ref GLOBAL: Arc<Global> = Arc::new(Global::new("wgpu", core::hub::IdentityManagerFactory));
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn wgpu_get_version() -> *const std::os::raw::c_char {
+    let version = env!("CARGO_PKG_VERSION");
+    let version_c = std::ffi::CString::new(version).unwrap();
+    let version_p = version_c.as_ptr();
+    std::mem::forget(version_c); // see https://thefullsnack.com/en/string-ffi-rust.html
+    return version_p;
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,10 +15,9 @@ lazy_static::lazy_static! {
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn wgpu_get_version() -> *const std::os::raw::c_char {
-    let version = env!("CARGO_PKG_VERSION");
-    let version_c = std::ffi::CString::new(version).unwrap();
-    let version_p = version_c.as_ptr();
-    std::mem::forget(version_c); // see https://thefullsnack.com/en/string-ffi-rust.html
-    return version_p;
+pub unsafe extern "C" fn wgpu_get_version() -> std::os::raw::c_uint {
+    let major: u32 = env!("CARGO_PKG_VERSION_MAJOR").parse().unwrap();
+    let minor: u32 = env!("CARGO_PKG_VERSION_MINOR").parse().unwrap();
+    let patch: u32 = env!("CARGO_PKG_VERSION_PATCH").parse().unwrap();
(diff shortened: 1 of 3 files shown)


almarklein:
What version of `wgpu-core` and `wgpu-types` will be used when I clone the `wgpu-native` repo and build it? Will it get the latest master from github, or will it download the latest 0.5 crate? Is there some way to control this?

Asking because with `wgpu-py` we're currently targeting 0.5.1, while the master of the wgpu-core repo is on 0.5.0, and the latest version is 0.5.3.

kvark:
This is how the dependencies work.
https://github.com/gfx-rs/wgpu-native/blob/master/Cargo.toml defines the path to look for. It points to git and says that the base 5.0 version is expected.
The exact git revisions are stored in "Cargo.lock" and updated if you do `cargo update -p wgpu-core` for example.

The reason `wgpu` repository has version 0.5.0 is because you are looking at the master branch. All the release fixes are back-ported to the v0.5 branch, and this is where the version bumps happen as well.

kvark:
If you want to stick strictly with wgpu-0.5, we'd want to also create a relesae branch in wgpu-native that we update with backwards-compatible fixes.

src/lib.rs
```diff
@@ -13,3 +13,12 @@ type Global = core::hub::Global<core::hub::IdentityManagerFactory>;
 lazy_static::lazy_static! {
     static ref GLOBAL: Arc<Global> = Arc::new(Global::new("wgpu", core::hub::IdentityManagerFactory));
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn wgpu_get_version() -> *const std::os::raw::c_char {
```
kvark, line 18: should this be a number instead?

almarklein: You mean pack three uint8's (major, minor, patch) into an uint32? That could work. Would you prefer that?

kvark: Yeah, that would allow to compare it with values like 0x501. I think I would prefer this.
@grovesNL do you have a preference?

almarklein:
Thanks for the info on how releases work. Will wgpu-native get patch releases too? And will these be matched to the version of wgpu-core?

kvark:
@almarklein what's going to happen in the near future is that wgpu-rs detaches itself from wgpu-native. It will have 3 backends:
  - wgpu-core
  - WebGPU
  - webgpu-native

That last choice is basically linking with a dynamic library that implements https://github.com/webgpu-native/webgpu-headers, querying entry points, etc. It could be wgpu-native, or Dawn, we want them used interchangeably.

So, strictly speaking, no Rust library or application would be linking to wgpu-native *directly* via Cargo. That means we don't have to have the versions match anything. It will, however, be very convenient for the users if they knew that, say, wgpu-rs-0.6.4 requires wgpu-native 0.6.2.

Going closer to the punch line. wgpu-core becomes an *internal* dependency of wgpu-native, so their versions don't have to match, strictly speaking.  We can still do patch releases, i.e. when we update the inner wgpu-core dependency, or add a new function.

almarklein:
Thanks. That sounds promising!

> We can still do patch releases, i.e. when we update the inner wgpu-core dependency, or add a new function.

👍  It would be great to build a specific version (rather than an arbitrary commit in master) of wgpu-native, and know what version of wgpu-core matches that. (Reproducibility and all that.)

src/lib.rs
kvark: is this still on your radar?

almarklein: You mean @grovesNL  or me?

kvark: I mean @almarklein  - to change this to u32

almarklein: Ah, I thought we were waiting for Josh' answer :) I'm on it.

almarklein: Done.

kvark:
bors retry

PR #15, opened 2020-04-29
Update to latest wgpu-core

almarklein:
Had a go at this locally to build a binary and figured I might as well submit a PR. 

I ran `cargo update -p wgpu-core`, made a few necessary adjustment, and ran `make ffi/wgpu.h`

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -289,6 +289,16 @@ typedef enum {
   WGPUVertexFormat_Int4 = 29,
 } WGPUVertexFormat;
 
+/**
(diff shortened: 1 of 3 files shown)


ffi/wgpu.h
```diff
@@ -728,6 +734,34 @@ typedef struct {
   uint32_t array_layer_count;
 } WGPUTextureViewDescriptor;
 
+
```
kvark, line 737: I wonder where these come from?

almarklein: I also thought it was strange that cbindgen would suddenly add those ...

PR #27, opened 2020-05-19
Update CI to publish binary releases on git version tag

almarklein:
This adds jobs to the Github action CI workflow, that build binaries suited for distribution. On commits that are tagged with `v...`, a matching release is made and the binaries (packed into zipfiles) are uploaded as assets. 

I basically took the Azure pipeline code from [wgpu-bin](https://github.com/Korijn/wgpu-bin), translated it to Github Actions, and weaved in into the existing CI. I deliberately made a new `release-build` job, so we can keep these builds separate from the `test-build` jobs.

I tested this on my own fork (Github Actions Just Work on forks!), I pushed a tag, and some minutes later [the releases appeared](https://github.com/almarklein/wgpu-native/releases).

So to publish a release, we must bump the version in cargo.toml, and then `git tag v..` and `git push` that tag. I considered detecting the version bump and creating a tag automatically, but IMO that feels like too much automation. Open for discussion though.

We have quite a lot of jobs now, some kind of duplicate. We could consider removing (some of) the `stable` test builds.


almarklein:
@kvark would you mind if I squash-merged this when we're ready? If I read it correctly, I can add co-authors then, and I'd like to attribute @Korijn, who authored wgpu-bin.

almarklein:
Two of the new builds pull a Docker image from quay.io, (by RedHat) which has been having issues all day. I assume that should normally be pretty stable :)

.github/workflows/ci.yml
```diff
@@ -9,20 +9,30 @@ on:
...
       fail-fast: false
       matrix:
-        name: [
-          MacOS Stable,
-          MacOS Nightly,
-          Ubuntu Stable,
-          Ubuntu Nightly,
-          Windows Stable,
-          Windows Nightly
-        ]
```
almarklein, line 25: Removed this, as it's not needed, I believe.

.github/workflows/ci.yml
```diff
@@ -96,43 +106,97 @@ jobs:
...
-          name: artifact
-          path: dist/wgpu-*.zip
 
-  publish:
-    if: contains(github.ref, 'tags/v')
-    needs: [build]
-    name: Publish release artifacts
-    runs-on: ubuntu-18.04
+  # Builds for binary releases.
+  # Also build on non-tag commits; it also functions as a test.
```
kvark, line 111: Do we have an idea of how much slower this would make our general CI? I.e. does using an image affect the build times?

Korijn:
In my experience, the rust build still largely dominates the time. The docker image is pretty small so its pulled in no time.

almarklein:
Looking at the reported times for each job, they are indeed very much comparable. The Windows builds take the longest, but they take about as long as the Windows builds that were already done. 

There are more jobs now, so potentially it can take more time if there's a queue for the pool of runners.

kvark:
@grovesNL would you be interested to look at this as well? I think it would be useful. Otherwise, we'll proceed.

.github/workflows/ci.yml
```diff
@@ -9,20 +9,30 @@ on:
...
+    - name: Create commit-sha file
+      env:
+        GITHUB_SHA: ${{ github.sha }}
+      run: |
+          echo $GITHUB_SHA > commit-sha
+    - name: Publish commit-sha
+      uses: actions/upload-artifact@v2
+      with:
+        path: commit-sha
+        name: dist
```
Korijn, line 27: Unlike wgpu-bin where we have to clone the source repo prior to building, here you can just create the commit-sha file in the publish job and avoid the preparation job completely. It will allow queueing the subsequent jobs earlier

.github/workflows/ci.yml
```diff
@@ -96,43 +106,97 @@ jobs:
...
-          release-url: ${{ steps.create_release.outputs.upload_url }}
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+    - uses: actions/checkout@v2
+    - name: Docker build
+      if: success() && matrix.OS_NAME == 'Linux'
+      env:
+        IMAGE: ${{ matrix.IMAGE }}
+        RUST_TOOLCHAIN: ${{ matrix.RUST_TOOLCHAIN }}
+      run: |
+          CID=$(docker create -t -w /tmp/wgpu-native -v $PWD:/tmp/src:ro quay.io/pypa/$IMAGE bash -c "\
```
Korijn, line 158: Don't you need to preface these bash scripts with `set -ex`? Or is that by default on github actions?

.github/workflows/ci.yml
grovesNL, line 111: I'd slightly prefer to only bother building tagged commits to avoid the potential for CI backlog, but we could start with this and see how it goes

kvark:
Sounds like we are good to merge this.
There are concerns voiced by @Korijn though that would need to be resolved.
bors delegate=Korjin

grovesNL:
(username typo)
bors delegate=Korijn

grovesNL:
cc @kvark ^

kvark:
bah
bors delegate=Korijn

kvark:
@grovesNL sorry about that! fixed now, you should have full control here.

Korijn:
No worries, it happens literally all the time! :)

Korijn:
@grovesNL just to make sure we are all clear on _why_ we're using that particular docker container:

* it helps create highly portable binaries for linux for reasons better explained here: https://www.python.org/dev/peps/pep-0513/
* the python package index only accepts binary packages that adhere to the linked PEP513 standard, so it was a requirement for wgpu-py

.github/workflows/ci.yml
almarklein: I'd assumed the build needed it, but `make package` already does that. We can thus omit this step completely, because the GH release page already reports the commit sha :)

.github/workflows/ci.yml
almarklein: The default is `set -e` (-like behavior), which can be turned off with `fail-fast: false`

Korijn: Ok, that's fine then

Korijn:
LGTM. @almarklein might want to give it one last full run at your fork?

almarklein:
@korijn That was a good call. Some fixes that I had earlier made to the release job had disappeared as I prepared this new branch. Anyway:
```
λ git tag v0.5.test5
λ git push origin v0.5.test5  # (origin being my fork)
```
-> https://github.com/almarklein/wgpu-native/releases/tag/v0.5.test5

kvark:
Strange, let me try
bors r=Korijn

kvark:
It shows some internal server error :/
Do we need to update bors.toml?

Korijn:
@almarklein any idea why build status is still pending?

kvark:
We'll need to merge this manually. That's ok. But let's also update "bors.toml" with the new task names.

kvark:
Also somewhat minor - there is a separate section in Readme describing the binary releases, which we should update according to this PR.

.github/workflows/ci.yml
Korijn: Turns out this was here for bors, see `bors.toml` :)

almarklein: Does it though? The jobs still have the same names. I don't think Bors checks this file or anything, it just listens for success/fail events with matching names, so I think it's fine. Unless I'm missing something.

kvark:
Are we all set here?

almarklein:
I think so. I made the release-builds only build for tags. The docker registry that the image is to be pulled from is having issues again, so I'm taking this as a sign that it's a liability.

Other than that, what up with that failing MacOS build suddenly?

Korijn:
> I think so. I made the release-builds only build for tags. The docker registry that the image is to be pulled from is having issues again, so I'm taking this as a sign that it's a liability.

quay.io is a very very common docker image registry, it's not normal for it to have issues like this.

> Other than that, what up with that failing MacOS build suddenly?

It fails when compiling dependency `ron`, probably an upstream issue?

```
   Compiling ron v0.5.1
thread 'rustc' panicked at 'index out of bounds: the len is 0 but the index is 0', src/librustc_mir_build/hair/pattern/_match.rs:1184:21

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports

note: rustc 1.45.0-nightly (9310e3bd4 2020-05-21) running on x86_64-apple-darwin

note: compiler flags: -C embed-bitcode=no -C debuginfo=2 --crate-type lib

note: some of the compiler flags provided by cargo are hidden

error: could not compile `gfx-backend-metal`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
##[error]Process completed with exit code 101.
```

kvark:
This is coming from gfx-backend-metal (and not RON), filed as https://github.com/rust-lang/rust/issues/72467
This PR needs to be merged manually, so it's not a blocker for you.

PR #28, opened 2020-05-19
Add 'wgpu_device_destroy_render_pipeline()' function

DavidPeicho:
Hi,

Nothing fancy, this PR simply forward the call `wgpu_device_destroy_render_pipeline()` to the `render_pipeline_destroy()` method in `wgpu-core`.


src/device.rs
```diff
@@ -325,6 +325,11 @@ pub extern "C" fn wgpu_device_create_render_pipeline(
     gfx_select!(device_id => GLOBAL.device_create_render_pipeline(device_id, desc, PhantomData))
 }
 
+#[no_mangle]
+pub extern "C" fn wgpu_device_destroy_render_pipeline(render_pipeline_id: id::RenderPipelineId) {
```
kvark, line 329: could we as well have `wgpu_device_destroy_compute_pipeline` since we are at it?

DavidPeicho:
@kvark added a few more missing. I also renamed the added destroy functions to match what you previously did (I removed the "device" part of the name which isn't necessary).

src/device.rs
```diff
@@ -326,7 +341,7 @@ pub extern "C" fn wgpu_device_create_render_pipeline(
 }
 
 #[no_mangle]
-pub extern "C" fn wgpu_device_destroy_render_pipeline(render_pipeline_id: id::RenderPipelineId) {
+pub extern "C" fn wgpu_destroy_render_pipeline(render_pipeline_id: id::RenderPipelineId) {
```
kvark, line 344: This is now inconsistent with the other destroy methods (others have `_destroy` at the end).

DavidPeicho: Oops, good catch. Let me fix that right now

kvark:
Please squash and also run `make` so that the C header is updated.

kvark:
Bah, nightly is all broken:
```
thread 'rustc' panicked at 'index out of bounds: the len is 0 but the index is 0', src/librustc_mir_build/hair/pattern/_match.rs:1184:21

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports

note: rustc 1.45.0-nightly (9310e3bd4 2020-05-21) running on x86_64-apple-darwin

note: compiler flags: -C embed-bitcode=no -C debuginfo=2 --crate-type lib

note: some of the compiler flags provided by cargo are hidden

error: could not compile `gfx-backend-metal`.
```

kvark:
Filed as https://github.com/rust-lang/rust/issues/72467

kvark:
bors retry

PR #31, opened 2020-05-24
Update copy functions, add write_texture

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/666
Notable API changes:
- `BufferCopyView` now includes a new `TextureDataLayout`.
- the `TextureCopyView::array_layer` is removed. Instead, the `origin.z` is treated as an array layer for non-3D textures.
- new `write_texture` method is here.

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -289,16 +289,6 @@ typedef enum {
   WGPUVertexFormat_Int4 = 29,
 } WGPUVertexFormat;
 
-/**
- * The internal enum mirrored from `BufferUsage`. The values don't have to match!
- */
-typedef struct WGPUBufferUse WGPUBufferUse;
-
-/**
- * The internal enum mirrored from `TextureUsage`. The values don't have to match!
- */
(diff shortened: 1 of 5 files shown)


ffi/wgpu.h
```diff
@@ -289,11 +289,15 @@ typedef enum {
...
 } WGPUVertexFormat;
 
+typedef struct WGPUBufferCopyView WGPUBufferCopyView;
+
 /**
  * The internal enum mirrored from `BufferUsage`. The values don't have to match!
  */
 typedef struct WGPUBufferUse WGPUBufferUse;
 
+typedef struct WGPUTextureCopyView WGPUTextureCopyView;
```
grovesNL, line 299: I don't think these changes are expected – maybe we need to whitelist `WGPUTextureCopyView`, `WGPUBufferCopyView`, and possibly others if they aren't automatically included (`WGPUOrigin3d`)?

kvark: fixing in https://github.com/gfx-rs/wgpu/pull/669

PR #34, opened 2020-06-02
Add Device/Limit Accessors and Anisotropy Extension

cwfitzgerald:
PR should be pretty self explanatory. Have not been able to test the examples yet, but that will be done before this merges.

As of right now, we have no protection from panics rippling up into c/c++ code. We might consider how to prevent this in a structured way. In my own codebase I use a proc macro to wrap all c interface functions with catch_unwind and a call to a global callback, but there are plenty of other options.

This is blocked until https://github.com/gfx-rs/wgpu/pull/675 gets brought into wgpu-native.

### DIFF
--- a/cbindgen.toml
+++ b/cbindgen.toml
@@ -9,15 +9,21 @@ typedef unsigned long long WGPUNonZeroU64;
 typedef unsigned long long WGPUOption_AdapterId;
 typedef unsigned long long WGPUOption_SurfaceId;
 typedef unsigned long long WGPUOption_TextureViewId;
+
+typedef struct WGPUChainedStruct WGPUChainedStruct;
 """
 include_version = true
 braces = "SameLine"
 line_length = 100
 tab_width = 2
 language = "C"
+style = "both"
 
 [export]
 prefix = "WGPU"
+include = [
+    "AnisotropicSamplerDescriptorExt"
+]
 exclude = [
 	"BufferUse",
 	"TextureUse",
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -12,6 +12,8 @@ typedef unsigned long long WGPUOption_AdapterId;
 typedef unsigned long long WGPUOption_SurfaceId;
 typedef unsigned long long WGPUOption_TextureViewId;
 
+typedef struct WGPUChainedStruct WGPUChainedStruct;
+
 
 #include <stdarg.h>
 #include <stdbool.h>
@@ -43,7 +45,13 @@ typedef unsigned long long WGPUOption_TextureViewId;
 
 #define WGPUMAX_VERTEX_BUFFERS 16
 
-typedef enum {
+typedef enum WGPUAddressMode {
+  WGPUAddressMode_ClampToEdge = 0,
+  WGPUAddressMode_Repeat = 1,
+  WGPUAddressMode_MirrorRepeat = 2,
+} WGPUAddressMode;
+
+typedef enum WGPUBindingType {
   WGPUBindingType_UniformBuffer = 0,
   WGPUBindingType_StorageBuffer = 1,
   WGPUBindingType_ReadonlyStorageBuffer = 2,
@@ -54,7 +62,7 @@ typedef enum {
   WGPUBindingType_WriteonlyStorageTexture = 7,
 } WGPUBindingType;
 
-typedef enum {
+typedef enum WGPUBlendFactor {
   WGPUBlendFactor_Zero = 0,
   WGPUBlendFactor_One = 1,
   WGPUBlendFactor_SrcColor = 2,
@@ -70,7 +78,7 @@ typedef enum {
   WGPUBlendFactor_OneMinusBlendColor = 12,
 } WGPUBlendFactor;
 
-typedef enum {
+typedef enum WGPUBlendOperation {
   WGPUBlendOperation_Add = 0,
   WGPUBlendOperation_Subtract = 1,
   WGPUBlendOperation_ReverseSubtract = 2,
@@ -78,14 +86,14 @@ typedef enum {
   WGPUBlendOperation_Max = 4,
 } WGPUBlendOperation;
 
-typedef enum {
+typedef enum WGPUBufferMapAsyncStatus {
   WGPUBufferMapAsyncStatus_Success,
   WGPUBufferMapAsyncStatus_Error,
   WGPUBufferMapAsyncStatus_Unknown,
   WGPUBufferMapAsyncStatus_ContextLost,
 } WGPUBufferMapAsyncStatus;
 
-typedef enum {
+typedef enum WGPUCompareFunction {
   WGPUCompareFunction_Undefined = 0,
   WGPUCompareFunction_Never = 1,
   WGPUCompareFunction_Less = 2,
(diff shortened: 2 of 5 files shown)


cbindgen.toml
```diff
@@ -18,6 +18,9 @@ language = "C"
 
 [export]
 prefix = "WGPU"
+include = [
+    "AnisotropicSamplerDescriptorExt"
```
kvark, line 22: why did you need to do this? Is the type skipped if not explicitly included?

ffi/wgpu.h
```diff
@@ -451,10 +480,12 @@ typedef WGPUNonZeroU64 WGPUId_Surface;
 
 typedef WGPUId_Surface WGPUSurfaceId;
 
+typedef uint64_t WGPUBufferSize;
+
 typedef struct {
   WGPUBufferId buffer;
   WGPUBufferAddress offset;
-  WGPUBufferAddress size;
+  WGPUBufferSize size;
```
kvark, line 488: unrelated to your PR, it's unfortunate that cbindgen ignored `WHOLE`. We'll need to define it or make sure cbindgen sees it

src/device.rs
```diff
@@ -213,12 +228,57 @@ pub extern "C" fn wgpu_texture_view_destroy(texture_view_id: id::TextureViewId)
     gfx_select!(texture_view_id => GLOBAL.texture_view_destroy(texture_view_id))
 }
 
+#[repr(C)]
+pub struct AnisotropicSamplerDescriptorExt {
+    pub next_in_chain: *const ChainedStruct,
```
kvark, line 233: Can we instead have `Option<&ChainedStruct>` here and everywhere else?

src/lib.rs
```diff
@@ -14,6 +14,82 @@ lazy_static::lazy_static! {
...
+///
+/// The result of these rules, and the fact that wgpu-native functions using it do not validate all these assumptions,
+/// using this macro is an indication that the function itself must be made unsafe.
+///
+/// # Notes
+///
+/// Given two or more extension structs of the same SType in the same chain, this macro will favor the latter most. There should
+/// not be more than one extension struct with the same SType in a chain anyway, so this behavior should be unproblematic.
+#[macro_export]
+macro_rules! follow_chain {
```
kvark, line 49: looking at this for a bit, I think it's really neat, well done!

cbindgen.toml
cwfitzgerald: Yeah it wasn't explicitly included, let me peruse the cbindgen docs to see if there's some flag we need to throw.

cwfitzgerald: There appears to be no way to force include all applicable structs. It either needs to be in this list, or used by a function.

ffi/wgpu.h
cwfitzgerald: I feel cbindgen config/setup in general needs a good working through, it mentions the instructions on how to build wgpu-remote :)

src/device.rs
cwfitzgerald: Should be all fixed. Worth a look inside the macro one more time to make sure I didn't accidentally any UB.

cbindgen.toml
kvark: Ah I see, that makes sense.

src/lib.rs
```diff
@@ -14,6 +14,82 @@ lazy_static::lazy_static! {
...
+    {
+        $(
+            let mut $stype: Option<&$ty> = None;
+        )*
+        let mut chain_opt: Option<&$crate::ChainedStruct> = $base.next_in_chain;
+        while let Some(next_in_chain) = chain_opt {
+            match next_in_chain.s_type {
+                $(
+                    $crate::SType::$stype => {
+                        let type_ptr: *const $ty = next_in_chain as *const $crate::ChainedStruct as _;
```
kvark, line 61: I think we need something like:
```rust
assert_eq!(0, (next_in_chain as *const _).align_offset(std::mem::align_of::<$ty>), "Chain structure is not properly aligned!");
```

src/lib.rs
```diff
@@ -14,6 +14,82 @@ lazy_static::lazy_static! {
...
+///
+/// `follow_chain!(map_thing_descriptor(base_c_descriptor, ThingDescriptorExtension1STypeValue => ThingDescriptorExtension1))`
+///
+/// # Safety
+///
+/// This macro does not use any internal unsafe blocks. The caller (or most likely the function) needs
+/// to be unsafe. The following constraints must be upheld for this to be valid:
+///
+/// - All pointers in the chain of next pointers must point to either null or a valid extension object
+/// - All non-null pointers must be sufficiently aligned such as they can be dereferenced as both a [`ChainedStruct`]
```
kvark, line 35: this line will then be possible to remove, since we panic otherwise (or can stop traversing if panic is not desired)

cwfitzgerald:
Should be all ready, just needs to be rebased once #35 merges.

cwfitzgerald:
Alright, this should be updated. Ostensibly nothing changed, but could you give it a once over, as enough stuff has happened in the rebase that something weird might have happened.

One thing I didn't catch in #35 is that `CLimits` needs to be called just plain `Limits` to match with the upstream header.

cwfitzgerald:
This seems to fail to build due to `WGPUChainedStruct` being a recursive struct, which this style of struct declaration doesn't support.

```rust
#[repr(C)]
pub struct ChainedStruct<'c> {
    next: Option<&'c ChainedStruct<'c>>,
    s_type: SType,
}
```

generates
```c
typedef struct {
  const WGPUChainedStruct *next;
  WGPUSType s_type;
} WGPUChainedStruct;
```

which doesn't work :(

kvark:
@cwfitzgerald let's switch the struct declaration type (see "style = " in [cbingen doc](https://github.com/eqrion/cbindgen/blob/master/docs.md)). I'm not clear on why the current one generates "type" style, given that the default should have been "both". Let's try to switch it to "tag"?

cwfitzgerald:
So setting it to both properly generates both, but still isnt' able to compile. It seems like cbindgen just can't generate recursive structures correctly, so we probably need to exclude that type and manually write the definition for it.

Korijn:
> So setting it to both properly generates both, but still isnt' able to compile. It seems like cbindgen just can't generate recursive structures correctly, so we probably need to exclude that type and manually write the definition for it.

Looks like we need to add `style="tag"` to the cbindgen config toml.

cwfitzgerald:
So that does indeed solve the issue with self-referential types (and I have pushed this), it gives rise to two different issues:

1) This breaks all of our C users. All references to wgpu tyles now need to be `struct Name` instead of just `Name`.
2) This makes the C++ api different from the C api (C++ is already broken for a couple reasons). C++ will allow you to just use `Name`, or use `struct Name` as a fallback.

kvark:
> This makes the C++ api different from the C api

Is this a problem though?

cwfitzgerald:
> Is this a problem though?

I definitely would consider it so, especially as it makes our interface different from webgpu-native.

Let me switch back to the old format and manually define that one recursive type for now, we need to properly revisit the cbindgen config, but that shouldn't block all PRs.

cwfitzgerald:
Alright, I managed to get it compiling with a simple forward declaration + both style. I think this is ready to merge.

ffi/wgpu.h
```diff
@@ -669,12 +696,31 @@ typedef struct {
   bool alpha_to_coverage_enabled;
 } WGPURenderPipelineDescriptor;
 
-typedef struct {
+typedef struct WGPUChainedStruct {
```
kvark, line 699: isn't this redefining the struct? I'm surprised this works :)

cwfitzgerald: Welcome to C! You first define the struct with a name, which puts it in the "struct" namespace, then typedef it to a normal name, which goes in the regular namespace. No idea how it interacts with C++ tbh, but thats a problem for another time.

cwfitzgerald:
bors r=kvark

ffi/wgpu.h
kvark: To clarify, I mean there is a previous definition:
```cpp
typedef struct WGPUChainedStruct WGPUChainedStruct;
```
I assumed that nothing is possible to define under this name afterwards, but here you are actually redefining the typedef. Still confused a bit, eh

cwfitzgerald: Same here tbh, I guessed this was okay because of how enums are declared with an unnamespaced name then the underlying type was also typedefed. C never ceases to confuse me.

PR #35, opened 2020-06-06
update bindings for wgpu #675 and #703

DevOrc:
Updates the bindings for https://github.com/gfx-rs/wgpu/pull/703 and https://github.com/gfx-rs/wgpu/pull/675

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -23,23 +23,25 @@ typedef unsigned long long WGPUOption_TextureViewId;
  */
 #define WGPUBIND_BUFFER_ALIGNMENT 256
 
+/**
+ * Buffer-Texture copies on command encoders have to have the `bytes_per_row`
+ * aligned to this number.
+ *
+ * This doesn't apply to `Queue::write_texture`.
+ */
+#define WGPUCOPY_BYTES_PER_ROW_ALIGNMENT 256
+
 #define WGPUDEFAULT_BIND_GROUPS 4
 
 #define WGPUDESIRED_NUM_FRAMES 3
 
-#define WGPUMAX_BIND_GROUPS 4
+#define WGPUMAX_ANISOTROPY 16
 
 #define WGPUMAX_COLOR_TARGETS 4
 
 #define WGPUMAX_MIP_LEVELS 16
 
 #define WGPUMAX_VERTEX_BUFFERS 16
-
-typedef enum {
-  WGPUAddressMode_ClampToEdge = 0,
-  WGPUAddressMode_Repeat = 1,
-  WGPUAddressMode_MirrorRepeat = 2,
-} WGPUAddressMode;
 
 typedef enum {
   WGPUBindingType_UniformBuffer = 0,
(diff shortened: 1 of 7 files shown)


DevOrc:
ill update this later today with the examples fixed

ffi/wgpu.h
```diff
@@ -298,6 +295,26 @@ typedef enum {
...
+ * default:
+ *
+ * ```ignore
+ * let limits = Limits {
+ *     max_bind_groups: 2,
+ *     ..Limits::default()
+ * }
+ * ```
+ */
+typedef struct WGPUNonExhaustive WGPUNonExhaustive;
```
kvark, line 312: we should just add types like this to the exclude list in `cbindgen.toml`

ffi/wgpu.h
```diff
@@ -655,19 +693,6 @@ typedef struct {
   bool alpha_to_coverage_enabled;
 } WGPURenderPipelineDescriptor;
 
-typedef struct {
```
kvark, line 658: now we need to declare this struct here, separately
the old struct turned to be Rust-only

cwfitzgerald: This is declared properly in #34 as it needs the pnext chain following. PR is fixing a wgpu PR before and after the one that #34 fixes, so we're kinda in this weird cyclical dependency state. I would say leave this as is for now until we can get #34 merged in right after this.

ffi/wgpu.h
```diff
@@ -298,6 +295,26 @@ typedef enum {
...
+ *     max_bind_groups: 2,
+ *     ..Limits::default()
+ * }
+ * ```
+ */
+typedef struct WGPUNonExhaustive WGPUNonExhaustive;
+
+typedef struct WGPUSamplerDescriptor WGPUSamplerDescriptor;
+
+typedef struct WGPUUnsafeExtensions WGPUUnsafeExtensions;
```
cwfitzgerald, line 316: This needs a C compatible analogue made, possibly manifesting as a raw boolean instead of this opaque type.

ffi/wgpu.h
kvark: ok, we can do that :)

kvark:
(CI and @cwfitzgerald concern are left to address before we go)

DevOrc:
So I think I fixed all of the issues except the compute example, which I need some help on.

For the new map_async logic, do you need to call [device_get_buffer_sub_data()](https://github.com/gfx-rs/wgpu/blob/b76c8481f8b6a2c2b50b737d966866071dd7a722/wgpu-core/src/device/mod.rs#L755) once the data is ready? The data used to be returned in the callback, but now that it is async I can't find a way to get the result.

kvark:
@DevOrc `device_get_buffer_sub_data` should not be used by wgpu-native, it's a semi-internal API... Useful for browser implementations.
Instead, after the callback for buffer returns, you can call `get_mapped_range` for a pointer to data. See [wgpu-rs example](https://github.com/gfx-rs/wgpu-rs/blob/45b79a21f19cf99d56fae1153efc9e841c8ef6bb/examples/hello-compute/main.rs#L118) for the reference.

DevOrc:
it should be squashed now. 

Thanks for all of the help!

cwfitzgerald:
Awesome!
bors r=kvark,cwfitzgerald

Issue #41, opened 2020-06-28
Obtaining AdapterInfo


DavidPeicho:
Hi,

It's currently not possible to obtain an `AdapterInfo` C struct. The issue is that the struct from `wgpu-core` contains a Rusty string. As @cwfitzgerald proposed, we could ask the user to provide a pointer filled by the call to `get_info`. Otherwise, I guess we would need a call to `destroy_adapter_info` that could free an allocated `CString`? First idea sounds better.

Thanks!

kvark:
Agreed, let's have the user providing the data.

DavidPeicho:
Can work on this except if @cwfitzgerald wants to do it

cwfitzgerald:
Be my guest! Always willing to help mentor on the matrix. https://matrix.to/#/#wgpu:matrix.org

DavidPeicho:
Should we close this one now that #43 is merged?

PR #43, opened 2020-07-11
adapter: add 'wgpu_adapter_get_info()'

DavidPeicho:
Hi,

This is an attempt to add the required feature(s) for #41.

Right now, the user is expected to provide a struct containing a preallocated `char*` for the name, something like that:

```C
char name[512] = {};
WGPUCAdapterInfo info;
info.name = name;
wgpu_adapter_get_info(adapter, &info);
```

We could also simply ask for the `char*` in the signature, and return a `WGPUCAdapterInfo`, something like:

```C
char name[512] = {};
WGPUCAdapterInfo info = wgpu_adapter_get_info(adapter, name);
```

It's nicer IMO, but sounds weird if we plan to put the name inside the struct as well... Maybe we users don't often need the name and we make a separate call? Or we make the struct `WGPUCAdapterInfo` without the name, good for me as well.

I had to make a custom `CDeviceType` that can be exposed in C. Do you think we should directly make the [DeviceType](https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/instance.rs#L279) `#[repr(u8)]` in wgpu-core?

Thanks!

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -1,8 +1,12 @@
 use crate::{follow_chain, ChainedStruct, SType, GLOBAL};
 
-use wgc::{device::HostMap, device::Label, gfx_select, hub::Token, id};
-use wgt::{BackendBit, DeviceDescriptor, Limits};
+use libc::c_char;
+use std::ptr;
 
+use wgc::{device::HostMap, device::Label, gfx_select, hub::Token, id, instance::DeviceType};
+use wgt::{Backend, BackendBit, DeviceDescriptor, Limits};
+
+use std::ffi::CString;
 use std::{marker::PhantomData, slice};
 
 pub type RequestAdapterCallback =
@@ -153,6 +157,48 @@ impl From<wgt::Limits> for CLimits {
             max_bind_groups: other.max_bind_groups,
         }
     }
+}
+
+#[repr(u8)]
+pub enum CDeviceType {
+    /// Other.
+    Other = 0,
+    /// Integrated GPU with shared CPU/GPU memory.
+    IntegratedGpu,
+    /// Discrete GPU with separate CPU/GPU memory.
+    DiscreteGpu,
+    /// Virtual / Hosted.
+    VirtualGpu,
+    /// Cpu / Software Rendering.
+    Cpu,
+}
+
+impl From<DeviceType> for CDeviceType {
+    fn from(other: DeviceType) -> Self {
+        match other {
+            DeviceType::Other => CDeviceType::Other,
+            DeviceType::IntegratedGpu => CDeviceType::IntegratedGpu,
+            DeviceType::DiscreteGpu => CDeviceType::DiscreteGpu,
+            DeviceType::VirtualGpu => CDeviceType::VirtualGpu,
+            DeviceType::Cpu => CDeviceType::Cpu,
+        }
+    }
+}
+
+#[repr(C)]
+pub struct CAdapterInfo {
+    /// Adapter name
+    pub name: *mut c_char,
+    /// Length of the adapter name
+    pub name_length: usize,
+    /// Vendor PCI id of the adapter
+    pub vendor: usize,
+    /// PCI id of the adapter
+    pub device: usize,
+    /// Type of device
+    pub device_type: CDeviceType,
+    /// Backend used for device
+    pub backend: Backend,
 }
 
 #[no_mangle]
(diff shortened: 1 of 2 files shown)


kvark:
> Do you think we should directly make the DeviceType #[repr(u8)] in wgpu-core?

We'd totally accept that

src/device.rs
```diff
@@ -155,6 +159,46 @@ impl From<wgt::Limits> for CLimits {
...
+            DeviceType::VirtualGpu => CDeviceType::VirtualGpu,
+            DeviceType::Cpu => CDeviceType::Cpu,
+        }
+    }
+}
+
+#[repr(C)]
+pub struct CAdapterInfo {
+    /// Adapter name
+    pub name: *mut c_char,
```
kvark, line 191: we need the user providing `name_length` variable, and then us using that much data and changing it

DavidPeicho:
Only issue I can see is that `wgpu-core` has changed a bit since **v0.0.5**. I made the PR for `DeviceType` in `wgpu-core` but we may need first to migrate `wgpu-native` to the latest version of core.

I am not sure exactly all what needs to be done, but I can give it a shot and help if needed. I saw that `RawPass` has been removed.

src/device.rs
```diff
@@ -535,3 +581,40 @@ pub extern "C" fn wgpu_buffer_get_mapped_range(
...
+/// The field `info.name` is expected to point to a pre-allocated memory
+/// location. This function is unsafe as there is no guarantee that the
+/// pointer is valid and big enough to hold the adapter name.
+#[no_mangle]
+pub unsafe extern "C" fn wgpu_adapter_get_info(
+    adapter_id: id::AdapterId,
+    info: &mut CAdapterInfo,
+) -> u8 {
+    let adapter_info = gfx_select!(adapter_id => GLOBAL.adapter_get_info(adapter_id));
+    let adapter_name = match CString::new(adapter_info.name) {
```
kvark, line 598: why would this fail? It seems to me that we could just `unwrap`

src/device.rs
```diff
@@ -535,3 +581,40 @@ pub extern "C" fn wgpu_buffer_get_mapped_range(
...
+
+    // Copies the string bytes owned into a the pre-allocated memory location
+    // pointed by `info.name`.
+    // NOTE: this is obviousy unsafe and the caller **must** ensure the
+    // memory is allocated.
+    let string_bytes = adapter_name.as_bytes_with_nul();
+    ptr::copy(
+        string_bytes.as_ptr(),
+        info.name as *mut u8,
+        std::cmp::max(info.name_length, string_bytes.len()),
```
kvark, line 616: this should be `min`, but also we need to make sure to add a terminating 0 if we aren't copying the whole `string_bytes`

kvark:
> Only issue I can see is that wgpu-core has changed a bit since v0.0.5. I made the PR for DeviceType in wgpu-core but we may need first to migrate wgpu-native to the latest version of core.

I see. You are right, at this moment we can't just get that. Let's have `DeviceType` in this crate separately, and then remove it once we update to `wgpu-core` that has your `repr(C)` fix.

src/device.rs
```diff
@@ -535,3 +581,40 @@ pub extern "C" fn wgpu_buffer_get_mapped_range(
...
+/// # Safety
+///
+/// The field `info.name` is expected to point to a pre-allocated memory
+/// location. This function is unsafe as there is no guarantee that the
+/// pointer is valid and big enough to hold the adapter name.
+#[no_mangle]
+pub unsafe extern "C" fn wgpu_adapter_get_info(
+    adapter_id: id::AdapterId,
+    info: &mut CAdapterInfo,
+) -> u8 {
```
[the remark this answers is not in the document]
DavidPeicho: This is kind of related to your next comment. If we unwrap the `CString::new` call, then we can just keep a void

src/device.rs
```diff
@@ -535,3 +581,35 @@ pub extern "C" fn wgpu_buffer_get_mapped_range(
...
+    info.backend = adapter_info.backend;
+
+    let string_bytes = adapter_name.as_bytes_with_nul();
+    let cpy_length = std::cmp::min(info.name_length, string_bytes.len());
+
+    // Copies the string bytes owned into a the pre-allocated memory location
+    // pointed by `info.name`.
+    // NOTE: this is obviousy unsafe and the caller **must** ensure the
+    // memory is allocated.
+    ptr::copy(string_bytes.as_ptr(), info.name as *mut u8, cpy_length - 1);
```
kvark, line 609: something needs to bail out if `cpy_length` is 0. Rust doesn't do underflow checks in Release.

DavidPeicho: I wasn't sure :D thanks for the precision!

PR #45, opened 2020-08-02
Moved development of the wgpu scopes wrapper to git

porky11:
It's now developed in a special group at GitLab instead of pijul


porky11:
@kvark 
Same, but more people know git, and pijul is still a bit buggy sometimes.
I might switch back, when the rework is done, but I'm not sure, if it's really still in work.

PR #47, opened 2020-09-06
Update to wgpu d90beda4e7f441cec1b18a994def2aceb68a8987

DevOrc:
This updates wgpu-native to [wgpu#d90beda4e7f441cec1b18a994def2aceb68a8987](https://github.com/gfx-rs/wgpu/commit/d90beda4e7f441cec1b18a994def2aceb68a8987)

__Key Changes__:

 - RenderPassId/ComputePassId changed to RenderPass* / ComputePass* 
 - Shader Module API Changed
- Debug Markers are implemented

__Concerns__: 
See comments below

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,6 +1,6 @@
 use crate::GLOBAL;
 
-pub use wgc::command::{bundle_ffi::*, compute_ffi::*, render_ffi::*};
+pub use wgc::command::{bundle_ffi::*, compute_ffi::*, render_ffi::*, ComputePass, RenderPass};
 
 use wgc::{gfx_select, id};
 
@@ -72,6 +72,31 @@ pub extern "C" fn wgpu_command_encoder_copy_texture_to_texture(
         copy_size))
 }
 
+type RenderPassColorAttachmentDescriptor =
+    wgt::RenderPassColorAttachmentDescriptorBase<wgc::id::TextureViewId>;
+type RenderPassDepthStencilAttachmentDescriptor =
+    wgt::RenderPassDepthStencilAttachmentDescriptorBase<wgc::id::TextureViewId>;
+
+#[repr(C)]
+pub struct RenderPassDescriptor<'a> {
+    color_attachments: &'a RenderPassColorAttachmentDescriptor,
+    color_attachments_length: usize,
+    depth_stencil_attachment: Option<&'a RenderPassDepthStencilAttachmentDescriptor>,
+}
+
+impl<'a> RenderPassDescriptor<'a> {
+    fn to_wgpu_type(&self) -> wgc::command::RenderPassDescriptor<'a> {
+        let color_attachments = unsafe {
+            std::slice::from_raw_parts(self.color_attachments, self.color_attachments_length)
+        };
+
+        wgc::command::RenderPassDescriptor {
+            color_attachments,
+            depth_stencil_attachment: self.depth_stencil_attachment,
+        }
+    }
+}
+
 /// # Safety
 ///
 /// This function is unsafe because improper use may lead to memory
@@ -80,9 +105,9 @@ pub extern "C" fn wgpu_command_encoder_copy_texture_to_texture(
 #[no_mangle]
 pub unsafe extern "C" fn wgpu_command_encoder_begin_render_pass(
     encoder_id: id::CommandEncoderId,
-    desc: &wgc::command::RenderPassDescriptor,
-) -> id::RenderPassId {
-    let pass = wgc::command::RawPass::new_render(encoder_id, desc);
+    desc: &RenderPassDescriptor,
+) -> *mut RenderPass {
+    let pass = wgc::command::RenderPass::new(encoder_id, desc.to_wgpu_type());
     Box::into_raw(Box::new(pass))
 }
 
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -29,7 +29,7 @@ impl OwnedLabel {
 }
 
 lazy_static::lazy_static! {
-    static ref GLOBAL: Arc<Global> = Arc::new(Global::new("wgpu", wgc::hub::IdentityManagerFactory));
+    static ref GLOBAL: Arc<Global> = Arc::new(Global::new("wgpu", wgc::hub::IdentityManagerFactory, wgt::BackendBit::PRIMARY));
```
DevOrc, line 32: Whats the best way to handle setting the backend bit?

src/device.rs
```diff
@@ -253,10 +261,10 @@ pub unsafe extern "C" fn wgpu_adapter_request_device(
...
 }
 
-#[no_mangle]
-pub extern "C" fn wgpu_adapter_extensions(adapter_id: id::AdapterId) -> wgt::Extensions {
-    gfx_select!(adapter_id => GLOBAL.adapter_extensions(adapter_id))
-}
+// #[no_mangle]
+// pub extern "C" fn wgpu_adapter_extensions(adapter_id: id::AdapterId) -> wgt::Extensions {
+//     gfx_select!(adapter_id => GLOBAL.adapter_extensions(adapter_id))
+// }
```
DevOrc, line 267: I couldn't find a replacement for this

src/device.rs
```diff
@@ -272,10 +280,10 @@ pub extern "C" fn wgpu_adapter_destroy(adapter_id: id::AdapterId) {
...
 }
 
-#[no_mangle]
-pub extern "C" fn wgpu_device_extensions(device_id: id::DeviceId) -> wgt::Extensions {
-    gfx_select!(device_id => GLOBAL.device_extensions(device_id))
-}
+// #[no_mangle]
+// pub extern "C" fn wgpu_device_extensions(device_id: id::DeviceId) -> wgt::Extensions {
+//     gfx_select!(device_id => GLOBAL.device_extensions(device_id))
+// }
```
DevOrc, line 286: Also couldn't find a replacement for this

src/device.rs
cwfitzgerald: The term "extensions" has been replaced with "features"

src/command.rs
```diff
@@ -72,6 +72,31 @@ pub extern "C" fn wgpu_command_encoder_copy_texture_to_texture(
...
+    wgt::RenderPassDepthStencilAttachmentDescriptorBase<wgc::id::TextureViewId>;
+
+#[repr(C)]
+pub struct RenderPassDescriptor<'a> {
+    color_attachments: &'a RenderPassColorAttachmentDescriptor,
+    color_attachments_length: usize,
+    depth_stencil_attachment: Option<&'a RenderPassDepthStencilAttachmentDescriptor>,
+}
+
+impl<'a> Into<wgc::command::RenderPassDescriptor<'a>> for &RenderPassDescriptor<'a> {
```
kvark, line 87: I think it's an antipattern to have `Into` implemented for `&Something`. When I see `xxx.into_yyy()` I expect stuff to be moved out, but with `&Something` it still gets borrowed, effectively implying `xxx.to_yyy()` semantics instead. Let's just make this a method on `RenderPassDescriptor`

src/command.rs
```diff
@@ -92,14 +117,16 @@ pub unsafe extern "C" fn wgpu_command_encoder_begin_render_pass(
 /// problems. For example, a double-free may occur if the function is called
 /// twice on the same raw pointer.
 #[no_mangle]
-pub unsafe extern "C" fn wgpu_render_pass_end_pass(pass_id: id::RenderPassId) {
-    let (pass_data, encoder_id) = Box::from_raw(pass_id).finish_render();
-    gfx_select!(encoder_id => GLOBAL.command_encoder_run_render_pass(encoder_id, &pass_data))
+pub unsafe extern "C" fn wgpu_render_pass_end_pass(pass: &mut RenderPass) {
```
kvark, line 120: this is UB to receive `&mut RenderPass` here, because we are producing a box from it, and then we have two different objects uniquely pointing to the same memory...

src/command.rs
```diff
@@ -111,18 +138,20 @@ pub unsafe extern "C" fn wgpu_render_pass_destroy(pass: id::RenderPassId) {
...
+) -> *mut ComputePass {
+    let pass = wgc::command::ComputePass::new(encoder_id);
     Box::into_raw(Box::new(pass))
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn wgpu_compute_pass_end_pass(pass_id: id::ComputePassId) {
-    let (pass_data, encoder_id) = Box::from_raw(pass_id).finish_compute();
-    gfx_select!(encoder_id => GLOBAL.command_encoder_run_compute_pass(encoder_id, &pass_data))
+pub unsafe extern "C" fn wgpu_compute_pass_end_pass(pass: &mut ComputePass) {
```
kvark, line 147: same here, we shouldn't take it as `&mut ComputePass`, but instead as `*mut ComputePass`

src/lib.rs
kvark: Good question! I think this should be come the `Instance` soon. We haven't discussed yet how the backends will be selected at this stage....

kvark (changes requested):
Amazing work, thank you!
It's also nice to see cbindgen transferring our comments to the header! Although, in the end we are going to use the shared header anyway, and it has no such comments...

DevOrc:
Ok everything should be solved now except for selecting the back ends

kvark:
Thank you!
Please mark the questions as "resolved" here once you address them.
bors r+

kvark (approved):

kvark:
bors retry

kvark:
Fixed CI issue in #48
bors retry

kvark:
@almarklein do you want a release right away, or first want to see if it works, and then give us a green light?

almarklein:
Let's do a release. If problems arise, we can fix them and release again. I'm not sure when I have the time to update the wrapper anyway, so if this is a good "snapshot", lets pint it :)

kvark:
@almarklein in terms of versions, it's likely that we'll need to be going v0.6, v0.7, v0.8, etc, independent of `wgpu`.
So the code needs Cargo.toml (and lock) updated, and README mentioning that the versions are unrelated to the upstream `wgpu`. Would you be good with that?

almarklein:
> Would you be good with that?

Sure, no problem!

kvark:
This happened :tada: 
https://github.com/gfx-rs/wgpu-native/releases/tag/v0.6.0

Issue #38, opened 2020-06-15
triangle example crashes when creating a new swapchain


radgeRayden:
As of the latest commit, the triangle example (and my program subsequently) crashes when creating a new swapchain, with the following backtrace: https://gist.github.com/radgeRayden/34173faca1f5b8d22ae086f9cd4dc89f

OS: Pop_OS 20.04
Graphics Adapter: Intel Corporation HD Graphics 5500 (rev 09)

kvark:
It looks like we are missing the `x11` feature enabled for `gfx-backend-vulkan` here. See how wgpu-rs [does that](https://github.com/gfx-rs/wgpu-rs/blob/0e2b612c5f33862b328b179e09ae0598cdb21fae/Cargo.toml#L51).

PR #49, opened 2020-09-12
Added X11 dependency to gfx-backend-vulkan

radgeRayden:
Fix for #38. It's copied from wgpu-rs, however it wouldn't work with gfx-backend-vulkan version 0.6 so I let it at version 0.5 since that has worked for me locally.


kvark (approved):
Good one, thank you!
(a bit sad now that we just released a version, maybe we should delete it?)
bors r+

PR #52, opened 2020-10-11
Update to wgpu 0.6.0

DevOrc:
Updates to wgpu-core 0.6.0. This is very much a WIP and many things still need to be implemented (textures, push constants), but I thought I would throw this up for comments. Also for error handling, anytime a result is an error, it just panics with a description of the function that panicked. 

 **Testing:** 
Triangle/Compute Examples

After I finish updating everything I will test it against my own stuff

### DIFF
--- a/cbindgen.toml
+++ b/cbindgen.toml
@@ -14,6 +14,7 @@ typedef unsigned long long WGPUOption_SamplerId;
 typedef unsigned long long WGPUOption_SurfaceId;
 typedef unsigned long long WGPUOption_TextureViewId;
 typedef unsigned long long WGPUOption_BufferSize;
+typedef unsigned long long WGPUOption_PipelineLayoutId;
 
 typedef struct WGPUChainedStruct WGPUChainedStruct;
 """
@@ -27,7 +28,8 @@ style = "both"
 [export]
 prefix = "WGPU"
 include = [
-    "AnisotropicSamplerDescriptorExt"
+    "AnisotropicSamplerDescriptorExt",
+    "SamplerBorderColor"
 ]
 exclude = [
 	"BufferUse",
@@ -38,8 +40,12 @@ exclude = [
 	"Option_SurfaceId",
 	"Option_BufferSize",
 	"Option_TextureViewId",
+	"Option_PipelineLayoutId",
 	"Option_NonZeroU32",
 	"Option_NonZeroU64",
+	"Option_SamplerBorderColor",
+	"Option_TextureViewDimension",
+	"Option_TextureFormat",
 	"wgpu_render_pass_finish",
 	"wgpu_compute_pass_finish",
 	"NonExhaustive",
@@ -52,6 +58,12 @@ exclude = [
 "TextureDescriptor_Label" = "TextureDescriptor"
 "TextureViewDescriptor_Label" = "TextureViewDescriptor"
 "SamplerDescriptor_Label" = "SamplerDescriptor"
+"BufferCopyViewC" = "BufferCopyView"
+"TextureCopyViewC" = "TextureCopyView"
+"RequestAdapterOptions_SurfaceId" = "RequestAdapterOptions"
+"Option_SamplerBorderColor" = "SamplerBorderColor"
+"Option_TextureViewDimension" = "TextureViewDimension"
+"Option_TextureFormat" = "TextureFormat"
 
 [parse]
 parse_deps = true
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,16 +1,29 @@
-use crate::GLOBAL;
+use crate::{Label, OwnedLabel, GLOBAL};
 
-pub use wgc::command::{bundle_ffi::*, compute_ffi::*, render_ffi::*, ComputePass, RenderPass};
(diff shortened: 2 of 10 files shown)


kvark:
Should this be marked as a "draft" until you finish?

DevOrc:
oops! meant to do that 😄

DevOrc:
This is hopefully finished now!

**Notes:**
- Updates wgpu-core to master branch ([rev 8ce2530b](https://github.com/gfx-rs/wgpu/tree/8ce2530b6963982ef9a58f7efac682d13ce4c508))
- Adds repr(C) versions for BufferCopyView, TextureCopyView, TextureViewDescriptor, CompareFunction, PipelineLayoutDescriptor, VertexBufferDescriptor, VertexStateDescriptor, RenderPipelineDescriptor, ComputePipelineDescriptor
- Adds error messages when wgpu-core returns and Err(_). 

**Testing**
wgpu-native and kgpu examples work

**cbingen bug note:**

The following pattern causes cbindgen to crash:
```rust
//In wgt
mod wgt {
    pub struct Foo<T> {
        foo: T,
    }
}

//In wgc
pub type Foo = wgt::Foo<u32>;

//In wgpu-native
mod native {
    #[repr(C)]
    pub struct Foo{
        foo: u32
    }
}
```
Basically, cbindgen will panic because it ignores the module name and therefore will have multiple types to export with the same name. This pattern is used by BufferCopyView and TextureCopyView. To work around it, the native versions of the struct are called BufferCopyViewC and TextureCopyViewC, and they are renamed when CBindgen generates the header. A solution would be to move TextureCopyView and BufferCopyView into wgpu-core and remove the type alias.

See https://github.com/eqrion/cbindgen/issues/573 for more details

kvark:
> To work around it, the native versions of the struct are called BufferCopyViewC and TextureCopyViewC

I think we used the `Raw` suffix for those previously. I don't mind either way, let's just try to be consistent here.

cbindgen.toml
```diff
@@ -27,7 +28,8 @@ style = "both"
 [export]
 prefix = "WGPU"
 include = [
-    "AnisotropicSamplerDescriptorExt"
+    "AnisotropicSamplerDescriptorExt",
+	"SamplerBorderColor"
```
kvark, line 32: tab detected :)

src/command.rs
```diff
@@ -28,67 +41,100 @@ pub extern "C" fn wgpu_command_encoder_copy_buffer_to_buffer(
...
+}
+
+#[repr(C)]
+#[derive(Clone)]
+pub struct BufferCopyViewC {
+    pub layout: wgt::TextureDataLayout,
+    pub buffer: id::BufferId,
+}
+
+impl Into<wgc::command::BufferCopyView> for BufferCopyViewC {
```
kvark, line 54: it would be simpler to just have this as `pub fn to_wgpu(&self) -> wgc::command::BufferCopyView`. since it would turn `xxx.cloned().into()` into `xxx.to_wgpu()`, which is much cleaner

src/command.rs
```diff
@@ -28,67 +41,100 @@ pub extern "C" fn wgpu_command_encoder_copy_buffer_to_buffer(
         destination,
         destination_offset,
         size))
+    .expect("Unable to copy buffer to buffer")
+}
+
+#[repr(C)]
+#[derive(Clone)]
+pub struct BufferCopyViewC {
```
kvark, line 49: would it help to slap `repr(C)` on the relevant `wgt` structs?

src/lib.rs
```diff
@@ -23,8 +26,11 @@ impl OwnedLabel {
             )
         })
     }
-    fn as_ref(&self) -> Option<&str> {
-        self.0.as_ref().map(|s| s.as_str())
+    fn ref_cow(&self) -> Option<Cow<str>> {
+        self.0.as_ref().map(|s| Cow::Borrowed(s.as_str()))
+    }
+    fn owned_cow<'a>(self) -> Option<Cow<'a, str>> {
```
kvark, line 32: should be named `into_cow`

src/lib.rs
```diff
@@ -23,8 +26,11 @@ impl OwnedLabel {
             )
         })
     }
-    fn as_ref(&self) -> Option<&str> {
-        self.0.as_ref().map(|s| s.as_str())
+    fn ref_cow(&self) -> Option<Cow<str>> {
```
kvark, line 29: we can call it `as_cow`

kvark (commented):
Amazing work here, thank you! A few notes before we proceed

src/command.rs
DevOrc: We can do that for now but we will eventually need to make a separate version here to match the [WebGPU headers](https://github.com/webgpu-native/webgpu-headers/blob/c761a96381b73f48d8e0416d0b7b4767e10a3860/webgpu.h#L750)

kvark: Right, so maybe we should try too hard to re-use it, I agree!

DevOrc:
Ok all of those recommendations should be fixed!

kvark (approved):
bors r+

PR #60, opened 2020-11-09
Rename TextureUsage::OUTPUT_ATTACHMENT to RENDER_ATTACHMENT

yzsolt:
Depends on https://github.com/gfx-rs/wgpu/pull/1017


ffi/wgpu.h
```diff
@@ -1558,9 +1558,9 @@ typedef uint32_t WGPUTextureUsage;
  */
 #define WGPUTextureUsage_STORAGE (uint32_t)8
 /**
- * Allows a texture to be a output attachment of a renderpass.
+ * Allows a texture to be an output attachment of a renderpass.
  */
-#define WGPUTextureUsage_OUTPUT_ATTACHMENT (uint32_t)16
+#define WGPUTextureUsage_RENDER_ATTACHMENT (uint32_t)16
```
kvark, line 1563: Did you change the file by hand?
I believe the `Cargo.lock` changes are missing from the PR

yzsolt: No, I used `make`, I was just intimidated by the huge diff in `Cargo.lock` at first, but now I see that it's mainly dependencies which were updated since the last `Cargo.lock` update. 
Double check please in case I messed up something!

kvark:
Generally, you don't need to update the whole cargo lock. Instead, just `cargo update -p wgpu-core` to update the relevant section. Also, let's make it squashed into 1 commit please!

kvark:
The GL backend support has been added to `wgpu`, so it has to be added here at least in some form. It's not the upstream changes that are missing, the error comes from this crate.

kvark:
Would you want to try enabling GL here? Not expecting you to test it, just make it build at least

yzsolt:
Yeah, sorry, by "upstream" I meant this crate, just from my viewpoint :)
Let's see if it builds now...

kvark (approved):
Thank yoU!
bors r+

kvark:
bors +

PR #65, opened 2020-12-25
Remove `C` prefix from some type names

grovesNL:
wgpu-native users shouldn't need to  be aware whether they're working with a separate C version of a type, so we should avoid using a prefix.


kvark (approved):
bors r+

PR #68, opened 2021-01-26
Add new capture example

Sineaggi:
Capture example is translated from the capture example in wgpu-rs
https://github.com/gfx-rs/wgpu-rs/tree/master/examples/capture

Currently this draft mr has two issues.
* The mr does not produce the same output as the upstream example
* The example does not write out a png

Although the code looks pretty similar to the code upstream, it produces results that are incorrect. Are these differences due to bugs in wgpu-native, or my translation?


kvark:
Ideally, for wgpu-native we wouldn't have a lot of examples, but instead host shared examples under `webgpu-headers` org/repo. But since we aren't there, adding an example here would definitely be useful for now.

> The mr does not produce the same output as the upstream example

There is a very easy way to find out the differences: record API traces from both paths (see https://github.com/gfx-rs/wgpu/wiki/Debugging-wgpu-Applications#tracing-infrastructure) and compare them.

Sineaggi:
@kvark 🤦 I have been staring at this for too long. The trace feature was immediately helpful. There was a missing call to _end_render_pass, which of course was being invoked in the rust example via the Drop trait.

Sineaggi:
The example now uses the calculated padding to correctly write out the example png.

examples/capture/main.c
```diff
@@ -0,0 +1,196 @@
...
+                                                              .label = NULL,
+                                                              0,
+                                                              wgpu_limits_default(),
+                                                              NULL}
+    );
+
+    BufferDimensions buffer_dimensions = buffer_dimensions_new(width, height);
+    WGPUBufferId output_buffer = wgpu_device_create_buffer(device,
+                                                           &(WGPUBufferDescriptor) {
+                                                                   .label = NULL,
```
kvark, line 99: could we use nicer formatting for C code?

examples/lodepng.c
```diff
@@ -0,0 +1,6464 @@
+/*
```
kvark, line 1: uuh, I'm not fully comfortable including this in the source.
Could we do something else instead? Like, link to `stb`, or hand-write the BMP output, or something like that (not involving a lot of code)?

kvark (commented):
Looks good! Just one concern.

examples/capture/main.c
Sineaggi: Yeah sorry about that, CLion has wild default formatting rules. I've used additional newlines to force better formatting.

examples/lodepng.c
Sineaggi: I just came across [this post](https://ibob.github.io/blog/2020/01/13/cmake-package-management/), is the current solution acceptable?

kvark: I think I have an idea of how to make this work without resorting to C libraries laying around. We could have another mini-crate as a part of this repository that's only job is to expose this:
```rust
extern "C" fn save_png(const char* path, const char* data);
```
The crate would be a single Rust file, non-publishable, and compiling to a static lib. It would just link to `png` crate and expose this FFI function. The capture example would then link with it and use the function.

Sineaggi: Ok so I've grabbed the code from the capture example in [wgpu-rs](https://github.com/gfx-rs/wgpu-rs/blob/master/examples/capture/main.rs#L143) and dropped it into an example helper library.

examples/helper/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+authors = ["Clayton Walker <clayton.m.walker@gmail.com>"]
+edition = "2018"
+
+[lib]
+crate-type = ["lib", "cdylib", "staticlib"]
+
+[dependencies]
+png = "0.16"
+libc = {version="0.2", features=[]}
```
kvark, line 12: should this just be `libc = "0.2"`?

examples/helper/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "helper"
+version = "0.1.0"
+authors = ["Clayton Walker <clayton.m.walker@gmail.com>"]
+edition = "2018"
```
kvark, line 5: let's have `publish = false` just to be sure

kvark (approved):
Looks great! Only a couple of nits noted, and let's convert it from the draft into a mergable PR

Sineaggi:
> Looks great! Only a couple of nits noted, and let's convert it from the draft into a mergable PR

Sounds good, would you like me to squash most of the commits?

examples/helper/Cargo.toml
Sineaggi: 👍

examples/helper/Cargo.toml
Sineaggi: I was using the same declaration as listed in the root wgpu-native [Cargo.toml](https://github.com/gfx-rs/wgpu-native/blob/master/Cargo.toml#L44). Fixed tho

kvark:
Yeah, let's squash some!

Sineaggi:
done

PR #71, opened 2021-02-09
Bump wgpu

Sineaggi:
First pass attempting to align wgpu-native with wgpu.

Going forward I'm wondering if the should focus should be making wgpu-native more aligned with wgpu, or start bringing it closer to webgpu-headers

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -835,35 +835,61 @@ impl<'a> ProgrammableStageDescriptor {
 }
 
 #[repr(C)]
-pub struct VertexBufferDescriptor {
-    pub stride: wgt::BufferAddress,
+pub struct VertexBufferLayout {
+    pub array_stride: wgt::BufferAddress,
     pub step_mode: wgt::InputStepMode,
-    pub attributes: *const wgt::VertexAttributeDescriptor,
-    pub attributes_length: usize,
+    pub attributes: *const wgt::VertexAttribute,
+    pub attributes_count: usize,
+}
+
+impl VertexBufferLayout {
+    unsafe fn into_wgpu(&self) -> wgc::pipeline::VertexBufferLayout {
+        wgc::pipeline::VertexBufferLayout {
+            array_stride: self.array_stride,
+            step_mode: self.step_mode,
+            attributes: Cow::Borrowed(slice::from_raw_parts(
+                self.attributes,
+                self.attributes_count,
+            )),
+        }
+    }
 }
 
 #[repr(C)]
-pub struct VertexStateDescriptor {
-    pub index_format: super::IndexFormat,
-    pub vertex_buffers: *const VertexBufferDescriptor,
-    pub vertex_buffers_length: usize,
+pub struct VertexState {
+    pub stage: ProgrammableStageDescriptor,
+    pub buffers: *const VertexBufferLayout,
+    pub buffer_count: usize,
+}
+
+#[repr(C)]
+pub struct FragmentState {
+    pub stage: ProgrammableStageDescriptor,
+    pub targets: *const wgt::ColorTargetState,
+    pub target_count: usize,
+}
+
+impl FragmentState {
+    unsafe fn into_wgpu(&self) -> wgc::pipeline::FragmentState {
+        wgc::pipeline::FragmentState {
+            stage: self.stage.into_wgpu(),
+            targets: Cow::Borrowed(slice::from_raw_parts(
+                self.targets,
+                self.target_count,
+            ))
+        }
+    }
 }
 
 #[repr(C)]
 pub struct RenderPipelineDescriptor {
     pub label: Label,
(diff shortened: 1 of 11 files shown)


kvark:
> Going forward I'm wondering if the should focus should be making wgpu-native more aligned with wgpu, or start bringing it closer to webgpu-headers

Bringing it closer to webgpu-headers is the real goal here.

almarklein:
@kvark it's been a while since the last release (with prebuild binaries). Would it be a good time to do a release after this PR merges?

src/device.rs
```diff
@@ -881,12 +881,33 @@ impl FragmentState {
     }
 }
 
+#[repr(C)]
+pub struct PrimitiveState {
+    pub topology: wgt::PrimitiveTopology,
+    pub strip_index_format: IndexFormat,
+    pub front_face: wgt::FrontFace,
+    pub cull_mode: wgt::CullMode,
```
kvark, line 889: this is `Option` now, actually, so if we pick up the latest we'll need to have this working with a local `enum CullMode`, just like we do with `IndexFormat`

kvark:
@almarklein sure thing! As long as we know that it works.
With wgpu-native it happens sometimes that some types just don't get properly exposed in `wgpu.h`, and we'd only know this once we start compiling code against it.

src/device.rs
```diff
@@ -835,78 +835,107 @@ impl<'a> ProgrammableStageDescriptor {
...
+    pub attributes: *const wgt::VertexAttribute,
+    pub attributes_count: usize,
+}
+
+impl VertexBufferLayout {
+    unsafe fn into_wgpu(&self) -> wgc::pipeline::VertexBufferLayout {
+        wgc::pipeline::VertexBufferLayout {
+            array_stride: self.array_stride,
+            step_mode: self.step_mode,
+            attributes: Cow::Borrowed(slice::from_raw_parts(
```
kvark, line 850: Here and everywhere else, we should be careful about getting a slice like this.
In Vulkan or OpenGL, if the count is 0, then the pointer should no be used. It can be garbage. In Rust, however, `slice::from_raw_parts` expects the pointer to not be garbage even if the size is 0.

I suggest avoiding the calls to `from_raw_parts` directly and instead using a helper:
```rust
unsafe fn make_slice<'a, T: 'a>(pointer: *const T, count: usize) -> &'a [T] {
    if count == 0 {
        &[]
    } else {
        slice::from_raw_parts(pointer, count)
    }
}
```

src/device.rs
```diff
@@ -835,78 +835,107 @@ impl<'a> ProgrammableStageDescriptor {
...
+
+#[repr(C)]
+pub struct FragmentState {
+    pub stage: ProgrammableStageDescriptor,
+    pub targets: *const wgt::ColorTargetState,
+    pub target_count: usize,
+}
+
+impl FragmentState {
+    unsafe fn into_wgpu(&self) -> wgc::pipeline::FragmentState {
```
kvark, line 873: same here - naming convention

src/device.rs
```diff
@@ -835,78 +835,107 @@ impl<'a> ProgrammableStageDescriptor {
...
+pub struct PrimitiveState {
+    pub topology: wgt::PrimitiveTopology,
+    pub strip_index_format: IndexFormat,
+    pub front_face: wgt::FrontFace,
+    pub cull_mode: wgt::CullMode,
+    pub polygon_mode: wgt::PolygonMode,
+}
+
+impl PrimitiveState {
+    fn into_wgpu(&self) -> wgt::PrimitiveState {
```
kvark, line 894: ditto

src/device.rs
```diff
@@ -835,78 +835,107 @@ impl<'a> ProgrammableStageDescriptor {
...
-        index_format: desc_base.vertex_state.index_format.into_wgpu(),
-        vertex_buffers: Cow::Borrowed(&vertex_buffers),
+    let buffers: Vec<_> = slice::from_raw_parts(
+        desc_base.vertex.buffers,
+        desc_base.vertex.buffer_count,
+    ).iter().map(|buffer| buffer.into_wgpu()).collect::<Vec<_>>();
+
+    let vertex = wgc::pipeline::VertexState {
+        stage: desc_base.vertex.stage.into_wgpu(),
+        buffers: Cow::Borrowed(&buffers),
```
kvark, line 928: should just do `Owned` since we own it

kvark (changes requested):

src/device.rs
Sineaggi: Makes sense, should I replace the 12 usages of `from_raw_parts` in device.rs?

kvark: yes, that would be good!

kvark (approved):
thank you!
bors r+

kvark:
bors delegate=Sineaggi

PR #72, opened 2021-02-11
Package DLL import lib on Windows

### LINKED ISSUE #57: Import *.lib for the Windows 32/64 packages is missing.
Hello, 

The import *.lib for Windows is missing from the Release packages ( @ https://github.com/gfx-rs/wgpu-native/releases)   Is the intention (for people that use the Release packages for Windows as opposed to compiling them themselves) to manually load the symbols for the DLL? Or is the import lib missing from the package?

Thanks in advance and for the great work, 
dmp

yzsolt:
Fixes #57.

Note: rustc/cargo emits `$lib_name.dll.lib` import libraries, which requires some tweaking by consumers, since the default expected name is `$lib_name.lib` (see [/IMPLIB documentation](https://docs.microsoft.com/en-us/cpp/build/reference/implib-name-import-library?view=msvc-160)). While we could rename it during packaging, I've decided not to, because:
 - it would cause a conflict if static libraries are put into the package too,
 - CMake similarly outputs `$libnamedll.lib` import libraries when building DLLs,
 - and once `wgpu-native` gets proper CMake integration, /IMPLIB can be overwritten easily (see [IMPORTED_IMPLIB documentation](https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED_IMPLIB.html)).

I've also removed the wildcards, according to our discussion on Discord (`wgpu-remote` has been removed, only `wgpu-native` is built now).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ FFI_DIR:=ffi
 BUILD_DIR:=build
 CLEAN_FFI_DIR:=
 CREATE_BUILD_DIR:=
+LIB_NAME:=wgpu_native
 
 WILDCARD_SOURCE:=$(wildcard src/*.rs)
 
@@ -32,10 +33,8 @@ endif
 ifeq ($(OS),Windows_NT)
 	LIB_EXTENSION=dll
 	OS_NAME=windows
-	ZIP_TOOL=7z
 else
 	UNAME_S:=$(shell uname -s)
-	ZIP_TOOL=zip
 	ifeq ($(UNAME_S),Linux)
 		LIB_EXTENSION=so
 		OS_NAME=linux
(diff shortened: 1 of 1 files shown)


kvark:
So we aren't packaging `.so` and `.dll` any more? Wouldn't these be useful for dynamic linking?

yzsolt:
> So we aren't packaging `.so` and `.dll` any more? Wouldn't these be useful for dynamic linking?

We still do. I've flipped the 7z/zip lines, maybe that's confusing a bit... The only real change is that now we also add `wgpu_native.dll.lib` to the package on Windows (required for implicit linking), anything else remains intact.

kvark:
Great!
bors r+

PR #75, opened 2021-02-20
Migrate pipeline creation to align with webgpu-headers

Sineaggi:
Edit(@kvark): solid step towards #6 

First effort trying to better align wgpu-native with webgpu-headers. Snake case was replaced with camel case, and the descriptor was rewritten to be closer to the structs that are defined in webgpu-headers.

Only the single function `wgpuDeviceCreateRenderPipeline` was changed in this pr to gauge interest. If merged, we'd be leaving wgpu-native in an inconsistent state.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -834,110 +834,287 @@ impl<'a> ProgrammableStageDescriptor {
     }
 }
 
-#[repr(C)]
-pub struct VertexBufferLayout {
-    pub array_stride: wgt::BufferAddress,
-    pub step_mode: wgt::InputStepMode,
-    pub attributes: *const wgt::VertexAttribute,
-    pub attributes_count: usize,
+#[repr(u32)]
+pub enum BlendOperation {
+    Add = 0,
+    Subtract = 1,
+    ReverseSubtract = 2,
+    Min = 3,
+    Max = 4,
 }
 
-impl VertexBufferLayout {
+impl BlendOperation {
+    fn to_wgpu(&self) -> wgt::BlendOperation {
+        match self {
+            BlendOperation::Add => wgt::BlendOperation::Add,
+            BlendOperation::Subtract => wgt::BlendOperation::Subtract,
+            BlendOperation::ReverseSubtract => wgt::BlendOperation::ReverseSubtract,
+            BlendOperation::Min => wgt::BlendOperation::Min,
+            BlendOperation::Max => wgt::BlendOperation::Max,
+        }
+    }
+}
+
+#[repr(u32)]
+pub enum BlendFactor {
+    Zero = 0,
+    One = 1,
+    SrcColor = 2,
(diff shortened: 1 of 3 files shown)


src/device.rs
```diff
@@ -834,110 +834,287 @@ impl<'a> ProgrammableStageDescriptor {
     }
 }
 
+#[repr(u32)]
+pub enum BlendOperation {
```
kvark, line 838: do we have to redefine these enums? they look exactly like the ones in `wgt`. Unless we are planning on adding the `Undefined = 0` variant to them, we could just keep using the `wgt` ones.

src/device.rs
```diff
@@ -834,110 +834,287 @@ impl<'a> ProgrammableStageDescriptor {
...
+            dst_factor: self.dstFactor.to_wgpu(),
+            operation: self.operation.to_wgpu(),
+        }
+    }
 }
 
-impl VertexBufferLayout {
+#[allow(non_snake_case)]
+#[repr(C)]
+pub struct ColorStateDescriptor<'c> {
```
kvark, line 915: neat lifetimes!

src/device.rs
```diff
@@ -834,110 +834,287 @@ impl<'a> ProgrammableStageDescriptor {
...
+}
+
+impl ColorStateDescriptor<'_> {
+    fn to_wgpu(&self) -> wgt::ColorTargetState {
+        wgt::ColorTargetState {
+            format: self.format,
+            alpha_blend: self.alphaBlend.to_wgpu(),
+            color_blend: self.colorBlend.to_wgpu(),
+            write_mask: self.writeMask,
+
```
kvark, line 930: weird newline

kvark (approved):

src/device.rs
Sineaggi: is `#[repr(C)]` different from `#[repr(u32)]`?

src/device.rs
Sineaggi: I don't know what these do in this context.

kvark: just allow us to have references instead of pointers to the chained structs, the chained structs are used to extend the basic functionality of WebGPU, for extensions and such

src/device.rs
kvark: Looks like they match - https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d3b8a1539dfe389d944dccfe8b7a6727

Sineaggi: So yes, we can avoid redefining these enums. I've replaced them with the ones in wgt.

src/device.rs
Sineaggi: cleaned up

kvark (approved):
Amazing work, thank you!
bors r+

PR #77, opened 2021-02-22
Migrate device methods to align with webgpu-headers

Sineaggi:
Second patch to align wgpu-native with webgpu-headers. This time focusing on the `wgpuDevice*` functions.

Largest change was to `wgpuDeviceCreateBindGroupLayout` which has been rebuilt to use the `*BindingLayout` structs.

Also wgpuDeviceCreateShaderModule now uses `follow_chain!` macro.


src/device.rs
```diff
@@ -221,38 +226,41 @@ pub struct CAdapterInfo {
     pub backend: Backend,
 }
 
+#[allow(non_snake_case)]
 #[repr(C)]
-pub struct CDeviceDescriptor {
+pub struct CDeviceDescriptor<'c> {
```
kvark, line 231: do we really need the `C` prefixes for these types? They are not colliding with `wgt::DeviceDescriptor`, since we aren't glob-importing `wgt`, so we could call them the same.

kvark (commented):
Great work! Just one note here

src/device.rs
kvark: See also - #78 as a radical solution to this

Sineaggi:
Closing as in favor of building off of the excellent work done in https://github.com/gfx-rs/wgpu-native/pull/85

PR #79, opened 2021-02-24
Update wgpu

Sineaggi:
Another bump to latest wgpu.

Needs the `"cross"` feature otherwise there's a nasty crash with an unhelpful backtrace.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -843,8 +843,8 @@ pub struct BlendDescriptor {
 }
 
 impl BlendDescriptor {
-    fn to_wgpu(&self) -> wgt::BlendState {
-        wgt::BlendState {
+    fn to_wgpu(&self) -> wgt::BlendComponent {
+        wgt::BlendComponent {
             src_factor: self.srcFactor,
             dst_factor: self.dstFactor,
             operation: self.operation,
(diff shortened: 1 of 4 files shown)


src/device.rs
```diff
@@ -866,8 +866,10 @@ impl ColorStateDescriptor<'_> {
     fn to_wgpu(&self) -> wgt::ColorTargetState {
         wgt::ColorTargetState {
             format: self.format,
-            alpha_blend: self.alphaBlend.to_wgpu(),
-            color_blend: self.colorBlend.to_wgpu(),
+            blend: Some(wgt::BlendState {
```
kvark, line 869: the headers weren't updated properly here. We want to make the `self.blend` thing, and have it nullable in webgpu-headers.
For now if we don't want to update the headers we could do:
```rust
blend: {
  let state = wgt::BlendState { ... }; //what you are doing now
  if state == wgt::BlendState::default() { None } else { Some(state) }
},
```

Sineaggi: Ok, I had checked in webgpu-headers to make sure they weren't nullable there first, before hardcoding the `Some`. If it's going to be nullable going forward, then this seems pretty reasonable.

kvark: I filed https://github.com/webgpu-native/webgpu-headers/issues/79 upstream

PR #84, opened 2021-03-26
add meson.build 

t18b219k:
This PR aims to add dependency resolving for meson user

If user want to build by meson it is not easy  due to lack of cargo support .
but this PR makes it easy i.e. user just write wgpu-native.wrap in subprojects

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,20 @@
+project('wgpu-native','c',version : '0.6.0')
+cargo = find_program('cargo')
+
+#build by cargo
+run_command(cargo,'build','--release')
+#mark cargo artifact to static library
+wgpu_native =static_library('wgpu_native',objects : 'target/release/libwgpu_native.a')
+
+dep_wgpu_native=declare_dependency(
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -0,0 +1,28 @@
+
+project('wgpu-native', 'c', version : '0.6.0', license : 'MPL 2.0')
+
+cargo = find_program('cargo')
+#determin whitch dir contain platform library
```
kvark, line 5: typos!

meson.build
```diff
@@ -0,0 +1,28 @@
+
+project('wgpu-native', 'c', version : '0.6.0', license : 'MPL 2.0')
```
kvark, line 2: are we going to need to bump this version every time we update Cargo.toml?

t18b219k: if we provide version information for meson we need it

meson.build
t18b219k: i fixed it

PR #87, opened 2021-04-19
Fix two error messages

almarklein:


kvark (approved):
Nice, thank you!
bors r+

PR #88, opened 2021-04-20
Fix struct references

Sineaggi:
Replaces 2 instances of 
```rust
descriptor: native::WGPURenderPassDescriptor
```
with
```rust
descriptor: &native::WGPURenderPassDescriptor
```

The code worked on Windows, but not on MacOS. Not sure about linux.


kvark (approved):
thank you!
bors r+

PR #89, opened 2021-04-20
Update gfx and naga to gfx-22

Sineaggi:
In order to support cases where enums in webgpu-headers and wgpu might diverge, I've added a new match for more direct mapping between two variants.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -330,7 +330,7 @@ dependencies = [
 [[package]]
 name = "gfx-auxil"
 version = "0.8.0"
-source = "git+https://github.com/gfx-rs/gfx?rev=d002952029facee7aafd52c94230057df6dd298f#d002952029facee7aafd52c94230057df6dd298f"
+source = "git+https://github.com/gfx-rs/gfx?rev=69991710a968e1e846148cdba425077e00b6febe#69991710a968e1e846148cdba425077e00b6febe"
 dependencies = [
  "fxhash",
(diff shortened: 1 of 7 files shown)


examples/capture/main.c
```diff
@@ -33,7 +33,7 @@ int main(
                 },
                 .maxBindGroups = 1,
                 .label = "Device",
-                .tracePath = NULL,
+                .tracePath = "build/",
```
kvark, line 36: leftover?

examples/triangle/main.c
```diff
@@ -123,7 +123,7 @@ int main()
                 },
                 .maxBindGroups = 1,
                 .label = "Device",
-                .tracePath = NULL,
+                .tracePath = "build/",
```
kvark, line 126: here also. is this intentional?

examples/capture/main.c
Sineaggi: was making it match the other one, maybe that was leftover? https://github.com/gfx-rs/wgpu-native/blob/master/examples/compute/main.c#L35

examples/triangle/main.c
Sineaggi: I don't mind making all of them NULL

kvark: let's make them NULL, yes

Sineaggi: Sounds good.

PR #93, opened 2021-04-30
Update readme to refer to wiki pages

almarklein:
Changes in wiki:
* Updated the getting-started page to include info on installing clang, remove reference to remote example, and added capture example.
* Added small page with some info on the binary releases.

In readme:
* Updated path to `webgpu.h`.
* Updated the link to the getting-started page.
* Added a link to the binary-releases page.

Should we also add something about the git submodule?


kvark (approved):
bors r+

kvark:
> Should we also add something about the git submodule?

we probably should, yes

almarklein:
> >  Should we also add something about the git submodule?
>
> we probably should, yes

done

PR #95, opened 2021-05-12
Bump to latest wgpu-core

almarklein:
* [x] Bump to wgpu-core (gets newer Naga too).
* [x] Some fixes to adjust to changes in wgpu-core.
* [x] Implement missing pieces in bind groups and bind group layouts (only buffers were supported so far).

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -155,11 +155,15 @@ pub unsafe extern "C" fn wgpuDeviceCreateBindGroupLayout(
         let ty = if is_texture {
             wgt::BindingType::Texture {
                 sample_type: match entry.texture.sampleType {
-                    native::WGPUTextureSampleType_Float => { wgt::TextureSampleType::Float {filterable: true} },
-                    native::WGPUTextureSampleType_UnfilterableFloat  => { wgt::TextureSampleType::Float {filterable: false} },
-                    native::WGPUTextureSampleType_Depth   => wgt::TextureSampleType::Depth,
-                    native::WGPUTextureSampleType_Sint   => wgt::TextureSampleType::Sint,
-                    native::WGPUTextureSampleType_Uint   => wgt::TextureSampleType::Uint,
+                    native::WGPUTextureSampleType_Float => {
+                        wgt::TextureSampleType::Float { filterable: true }
+                    }
+                    native::WGPUTextureSampleType_UnfilterableFloat => {
+                        wgt::TextureSampleType::Float { filterable: false }
+                    }
+                    native::WGPUTextureSampleType_Depth => wgt::TextureSampleType::Depth,
+                    native::WGPUTextureSampleType_Sint => wgt::TextureSampleType::Sint,
+                    native::WGPUTextureSampleType_Uint => wgt::TextureSampleType::Uint,
                     x => panic!("Unknown texture SampleType: {}", x),
                 },
                 view_dimension: match entry.texture.viewDimension {
@@ -167,7 +171,9 @@ pub unsafe extern "C" fn wgpuDeviceCreateBindGroupLayout(
                     native::WGPUTextureViewDimension_2D => wgt::TextureViewDimension::D2,
                     native::WGPUTextureViewDimension_2DArray => wgt::TextureViewDimension::D2Array,
                     native::WGPUTextureViewDimension_Cube => wgt::TextureViewDimension::Cube,
-                    native::WGPUTextureViewDimension_CubeArray => wgt::TextureViewDimension::CubeArray,
+                    native::WGPUTextureViewDimension_CubeArray => {
+                        wgt::TextureViewDimension::CubeArray
+                    }
                     native::WGPUTextureViewDimension_3D => wgt::TextureViewDimension::D3,
                     x => panic!("Unknown texture ViewDimension: {}", x),
                 },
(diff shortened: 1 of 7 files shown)


src/device.rs
```diff
@@ -202,8 +235,7 @@ pub unsafe extern "C" fn wgpuDeviceCreateBindGroup(
     let mut entries = Vec::new();
 
     for entry in make_slice(descriptor.entries, descriptor.entryCount as usize) {
-        entries.push(
-            //TODO: Not all things are buffers!
+        let wgc_entry = if entry.buffer {
```
almarklein, line 238: I need some help here. The `entry.buffer` is a `Id<wgpu_core::resource::Buffer<wgpu_core::backend::Empty>>`, but can be Null. How to check that here, I can't figure it out :P

kvark: if it can be null, then the Rust type needs to be `Option<Id<...>>`

almarklein: Ok ... so to be clear, we have to go from [this C struct](https://github.com/webgpu-native/webgpu-headers/blob/8b8c2a77dd5a94a64d987b664e2ec320e34a3dc5/webgpu.h#L521-L528) to [this Rust struct](https://docs.rs/wgpu-core/0.8.1/wgpu_core/binding_model/struct.BindGroupEntry.html). How/where should we define that the id's for `buffer`, `sampler`  and `textureView` can be null/zero?

kvark: Good question! Gotta see what fine-grained options `bindgen` might have...
In the worst case, [this mapping](https://github.com/gfx-rs/wgpu-native/blob/b4055903d75a2d95fcf71adc8f5ec13b48625e49/build.rs#L9) would have everything as `Option<Id<..>>`

almarklein: But then all id's would be `Option<Id>` wouldn't they? Ideally it would only apply to the fields of this specific C struct. What about comparing it to something like `Id(0)`?

almarklein: An `Id` is a [wrapper](https://github.com/gfx-rs/wgpu/blob/ff5e626bfd04e33f5b5c9828b27f262308a2dd0d/wgpu-core/src/id.rs) around a `NonZeroU64`. I've tried to find a way to check that that value is zero, but so far I've not found a way that the Rust compiler allows. These id-fields in the C-struct are supposed to be all zero except one. But can they even be zero if the type is `NonZeroU64`? Not sure how to proceed t.b.h.

kvark: `NonZero` can't be zero by definition, you can't check that (because the compiler just assumes that unconditionally).
So if your FFI expects a NonZero but you pass 0 there - that's classy UB. That is to say, this is UB in a lot of other cases too, e.g. when the ID is totally random. We webgpu implementation can be backing IDs by pointers, in which case random data leads to access violations...

So the safety guarantees provided by webgpu only kick in if we assume that all of the input IDs are valid (or 0 where the spec explicitly allows NULL). Therefore, it's totally fine for us to expect a NonZero (or `Id`) in most places, but these bind group entries must be an exception. We need to convince bindgen that in the case of this struct (only), it should use `Option<Id<>>` instead of `Id<>`.

almarklein: > We need to convince bindgen that in the case of this struct (only), it should use `Option<Id<>>` instead of `Id<>`.

Thanks, that makes sense. Do you have any tips / starting points to help a Rust novice like me on the way to implement this?

kvark: Here is an idea. `bindgen` supports blacklisting types, and also allows one to generate a header. So this particular structure `WGPUBindGroupEntry` could be black-listed, and we'd just write down our own version of it in the header.

See how the IDs are [blacklisted](https://github.com/gfx-rs/wgpu-native/blob/b4055903d75a2d95fcf71adc8f5ec13b48625e49/build.rs#L46-L48) and replaced, similarly.

almarklein: Thanks! It works!

almarklein:
This is ready for review. Thanks for the help!

almarklein:
There may be more regressions that I'll run into (e.g. #96), but can fit those in a new PR.

I updated to the latest wgpu-core where naga was bumped to 25. In that PR you noted:

> this PR doesn't currently work because Metal's backend is now required use a sizes buffer when dynamically sized arrays are present. One of the internal tests here uses them.

What exactly are the implications of that?

almarklein:
Also fixed some (but not all) of the missing functions reported in #96

src/device.rs
```diff
@@ -153,11 +154,73 @@ pub unsafe extern "C" fn wgpuDeviceCreateBindGroupLayout(
...
+                native::WGPUSamplerBindingType_Filtering => wgt::BindingType::Sampler {
+                    filtering: true,
+                    comparison: false,
+                },
+                native::WGPUSamplerBindingType_NonFiltering => wgt::BindingType::Sampler {
+                    filtering: false,
+                    comparison: false,
+                },
+                native::WGPUSamplerBindingType_Comparison => wgt::BindingType::Sampler {
+                    filtering: false,
```
kvark, line 194: should probably be `true` here

src/device.rs
```diff
@@ -308,6 +384,26 @@ pub unsafe extern "C" fn wgpuQueueWriteBuffer(
         .expect("Unable to write buffer")
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn wgpuQueueWriteTexture(
+    queue: id::QueueId,
+    destination: native::WGPUImageCopyTexture,
+    data: *const u8, // TODO: Check - this might not follow the header
```
kvark, line 391: header has `*const c_void` basically. Probably fine then?

kvark (approved):

kvark:
Great stuff, thank you!
bors r+

PR #97, opened 2021-05-17
Fix triangle example (do not use empty bind groups)

almarklein:
The triangle example in wgpu-py started failing for me recently. Indeed the example in wgpu-native fails too now (for my system at least). This fixes it.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -299,9 +299,6 @@ pub unsafe extern "C" fn wgpuDeviceCreateBindGroup(
         layout: descriptor.layout,
         entries: Cow::Borrowed(&entries),
     };
-    if entries.len() == 0 {
-        panic!("Bind groups with zero entries not allowed.");
-    }
     check_error(gfx_select!(device => GLOBAL.device_create_bind_group(device, &desc, PhantomData)))
 }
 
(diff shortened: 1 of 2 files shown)


src/device.rs
```diff
@@ -299,6 +299,9 @@ pub unsafe extern "C" fn wgpuDeviceCreateBindGroup(
         layout: descriptor.layout,
         entries: Cow::Borrowed(&entries),
     };
+    if entries.len() == 0 {
+        panic!("Bind groups with zero entries not allowed.");
+    }
```
almarklein, line 304: Otherwise, without validation layers, I suppose you'd not see an error but simply crash.

almarklein: But its a bit ahead of a decision on https://github.com/gfx-rs/wgpu/issues/240

almarklein:
Maybe await https://github.com/zakarumych/gpu-descriptor/issues/14

src/device.rs
kvark: this shouldn't be handled in wgpu-native

almarklein: Agreed. Will remove it.

Do you think the changes to the triangle example should still be applied, or do you prefer closing this PR, since it looks like the problem will be resolved?

kvark: I think it's good to still fix the example. Empty bind groups aren't very useful :)

almarklein: Ok,. I removed the check.

kvark (approved):
thanks!
bors r+

PR #98, opened 2021-05-19
Add a few missing functions

almarklein:
This adds a few more missing functions. Enough to again have every feature that wgpu-py is currently covering and making all of wgpu-py's tests pass. Also bumped the wgpu-core commit sha.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -504,7 +504,37 @@ pub unsafe extern "C" fn wgpuDeviceCreateRenderPipeline(
             polygon_mode: wgt::PolygonMode::Fill,
             conservative: false,
         },
-        depth_stencil: descriptor.depthStencil.as_ref().map(|_| unimplemented!()),
+        depth_stencil: descriptor
+            .depthStencil
+            .as_ref()
+            .map(|desc| wgt::DepthStencilState {
+                format: map_texture_format(desc.format)
+                    .expect("Texture format must be defined in DepthStencilState"),
+                depth_write_enabled: desc.depthWriteEnabled,
+                depth_compare: map_compare_function(desc.depthCompare).unwrap(),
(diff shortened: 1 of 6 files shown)


src/device.rs
```diff
@@ -504,7 +504,37 @@ pub unsafe extern "C" fn wgpuDeviceCreateRenderPipeline(
...
+        depth_stencil: descriptor
+            .depthStencil
+            .as_ref()
+            .map(|desc| wgt::DepthStencilState {
+                format: map_texture_format(desc.format)
+                    .expect("Texture format must be defined in DepthStencilState"),
+                depth_write_enabled: desc.depthWriteEnabled,
+                depth_compare: map_compare_function(desc.depthCompare).unwrap(),
+                stencil: wgt::StencilState {
+                    front: wgt::StencilFaceState {
```
kvark, line 516: can we have a `map_stencil_face` helper to avoid repeating ourselves?

src/device.rs
```diff
@@ -766,6 +796,19 @@ map_enum!(
     Min,
     Max
 );
+map_enum!(
```
kvark, line 799: In other parts of gfx/wgpu ecosystem, we tend to have a specific module (like `conv.rs`) to convert things. Maybe it is useful here too, so that we can make `device.rs` lighter. Up to you :)

kvark (commented):
Nice work! Just one note

src/device.rs
almarklein: Yes!

almarklein:
Both done!

almarklein:
I also moved the `map_x` enums and functions from `command.rs`  there.

kvark (approved):
wonderful!
bors r+

PR #100, opened 2021-05-19
Add surfaceGetPreferredFormat

DevOrc:
Adds wgpuSurfaceGetPreferredFormat and updates the triangle example. 

Right now, `map_enum!` only converts native types to Rust types, so I had to manually map the `wgt::TextureFormat` to `native::WGPUTextureFormat`. In the future, it might be prudent to rework `map_enum!` to create mapping functions for both directions.


kvark (approved):
bors r+

PR #102, opened 2021-05-20
Allow setting backendbits when creating an adapter

### LINKED ISSUE #101: Ability to select backend
It seems that depending on the settings in my Nvida Conrol Panel (whether I select Integrated Graphics or the NVidea GPU), wgpu will use either the Vulkan or DX12 backend. With the DX12 backend I run into crashes for some code, which I should probably look into at some point, but it would be nice to just force the use of the Vulkan backend for now. Also because (in my understanding) the Vulkan backend is more mature.

Is this something that could be added to wgpu-native, e.g. in the form of an additional function `wgpuInstanceRequestAdapterSelectBackend` that has an extra argument for the backend bits? This would make the API deviate from webgpu, but its something that might be worthy for desktop use (similar to e.g. the tracing functionality). I'd be happy to prepare a PR for this.

almarklein:
Closes #101

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -182,11 +182,11 @@ pub fn map_origin3d(native: &native::WGPUOrigin3D) -> wgt::Origin3d {
 pub fn map_adapter_options<'a>(
     options: &native::WGPURequestAdapterOptions,
     extras: Option<&native::WGPUAdapterExtras>,
-) -> (Option<id::SurfaceId>, u32) {
+) -> (Option<id::SurfaceId>, native::WGPUBackendType) {
     if let Some(extras) = extras {
-        (options.compatibleSurface, extras.backend as u32)
+        (options.compatibleSurface, extras.backend)
     } else {
-        (options.compatibleSurface, 0u32)
+        (options.compatibleSurface, native::WGPUBackendType_Null)
     }
 }
 
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -184,9 +184,9 @@ pub fn map_adapter_options<'a>(
     extras: Option<&native::WGPUAdapterExtras>,
 ) -> (Option<id::SurfaceId>, u32) {
     if let Some(extras) = extras {
-        (Some(options.compatibleSurface), extras.backendBits as u32)
+        (options.compatibleSurface, extras.backend as u32)
     } else {
-        (Some(options.compatibleSurface), 0u32)
+        (options.compatibleSurface, 0u32)
     }
 }
 
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -182,11 +182,11 @@ pub fn map_origin3d(native: &native::WGPUOrigin3D) -> wgt::Origin3d {
 pub fn map_adapter_options<'a>(
     options: &native::WGPURequestAdapterOptions,
     extras: Option<&native::WGPUAdapterExtras>,
-) -> (Option<id::SurfaceId>, i32) {
+) -> (Option<id::SurfaceId>, u32) {
     if let Some(extras) = extras {
-        (Some(options.compatibleSurface), extras.backendBits)
+        (Some(options.compatibleSurface), extras.backendBits as u32)
     } else {
-        (Some(options.compatibleSurface), 0i32)
+        (Some(options.compatibleSurface), 0u32)
     }
 }
 
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -1,6 +1,6 @@
 use crate::{make_slice, map_enum, native, Label, OwnedLabel};
 use std::{borrow::Cow, ffi::CStr, num::NonZeroU32};
-use wgc::pipeline::ShaderModuleSource;
+use wgc::{id, pipeline::ShaderModuleSource};
 
 map_enum!(
     map_load_op,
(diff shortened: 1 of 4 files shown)


src/conv.rs
```diff
@@ -179,6 +179,17 @@ pub fn map_origin3d(native: &native::WGPUOrigin3D) -> wgt::Origin3d {
     }
 }
 
+pub fn map_adapter_options<'a>(
+    options: &native::WGPURequestAdapterOptions,
+    extras: Option<&native::WGPUAdapterExtras>,
+) -> (Option<id::SurfaceId>, i32) {
+    if let Some(extras) = extras {
+        (Some(options.compatibleSurface), extras.backendBits)
```
almarklein, line 187: I'm not so sure about the `Some` here. The `compatibleSurface` can be null, right?

src/device.rs
```diff
@@ -12,19 +12,38 @@ use wgc::{gfx_select, id};
 #[no_mangle]
 pub unsafe extern "C" fn wgpuInstanceRequestAdapter(
     _: native::WGPUInstance,
-    options: *const native::WGPURequestAdapterOptions,
+    options: &native::WGPURequestAdapterOptions,
     callback: native::WGPURequestAdapterCallback,
     userdata: *mut std::os::raw::c_void,
 ) {
-    let compatible_surface: Option<id::SurfaceId> =
-        options.as_ref().map(|options| options.compatibleSurface);
```
almarklein, line 20: This shows how the `Option<id::SurfaceId>` was obtained first. But I have trouble following how this allowed the surface to be optional.

src/conv.rs
```diff
@@ -179,6 +179,17 @@ pub fn map_origin3d(native: &native::WGPUOrigin3D) -> wgt::Origin3d {
     }
 }
 
+pub fn map_adapter_options<'a>(
+    options: &native::WGPURequestAdapterOptions,
+    extras: Option<&native::WGPUAdapterExtras>,
+) -> (Option<id::SurfaceId>, u32) {
+    if let Some(extras) = extras {
+        (Some(options.compatibleSurface), extras.backendBits as u32)
```
kvark, line 187: why is this always `Some`?

src/device.rs
kvark: this was incorrect :/

src/device.rs
```diff
@@ -12,19 +12,38 @@ use wgc::{gfx_select, id};
...
+    let (compatible_surface, given_backend_bits) = follow_chain!(
+        map_adapter_options(options,
+        WGPUSType_AdapterExtras => native::WGPUAdapterExtras)
+    );
+    // Convert backend bits. We need to initialize with *something*, thus the xor at the end.
+    let mut backend_bits = wgt::BackendBit::VULKAN;
```
kvark, line 24: why not initialize it with 0?

src/device.rs
```diff
@@ -12,19 +12,38 @@ use wgc::{gfx_select, id};
...
+    if given_backend_bits & native::WGPUBackendBits_Metal as u32 > 0 {
```
kvark, line 25: where are we defaulting to PRIMARY? We should be defaulting if the user doesn't pass the extra structure

src/device.rs
almarklein: I don't know how. There is `None` field in `wgt::BackendBit`. And if I initialize to `0u32` I get `no implementation for u32 |= BackendBit`.

src/conv.rs
almarklein: Because Rust thinks `options.compatibleSurface` is `SurfaceId`, not `Option<SurfaceId>`. I suspect we need to apply similar trickery as earlier to tell bindgen?

It *does* work now. I.e. I set `options.comatibleSurface` to null (when calling the C-api). So maybe the `SurfaceId` value is somehow checked down the road? Maybe it is just ignored if it does not match an existing surface?

src/device.rs
almarklein: Fixed.

almarklein:
I just saw there already is a `WGPUBackendType` enum in `webgpu.h`, but it can't be used as a bit field.

src/conv.rs
kvark: Yes, definitely, we need a similar trickery every time there is an Option<Id>

src/device.rs
kvark: `wgt::BackendBit::empty()`

almarklein:
Ready for review again. I updated `build.rs` to tell Rust that `compatibleSurface` is optional.

I also changed this to make use of the existing `WGPUBackendType` enum. This results in less code, but as a consequence the user-selected backend is not a bitfield but one specific backend. Let me know if you prefer the previous solution with an actual bitfield.

The answer to that question may also depend on https://github.com/gfx-rs/wgpu-native/issues/101#issuecomment-845714542 If the automatic backend selection would be flawless, the need for overriding the behavior would be confined mostly to testing.

kvark:
At the end of the day, we'll follow the solution (of backend selection) that is figured out in webgpu-headers.
For now, your approach is good.

almarklein:
Oh, I was not expecting backend selection to be part of the webgpu API :)

kvark:
It's not a part of WebGPU upstream API, but it could be a part of webgpu-native.

PR #103, opened 2021-05-21
Implement stub wgpuAdapterGetProperties

almarklein:
Would be nice to be able to get some feedback on what hardware and backend are selected.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -64,16 +64,30 @@ pub unsafe extern "C" fn wgpuAdapterGetProperties(
     adapter: id::AdapterId,
     properties: &mut native::WGPUAdapterProperties,
 ) {
-    //let maybe_props: Result<wgt::AdapterInfo, wgc::instance::InvalidAdapter> = GLOBAL.adapter_get_info(adapter);
-
     let maybe_props = gfx_select!(adapter => GLOBAL.adapter_get_info(adapter));
     match maybe_props {
         Ok(props) => {
-            //properties.name = std::ffi::CString::new(props.name).unwrap().as_ptr();
+            //let name = std::ffi::CString::new(props.name);
+            //store_object_globally(name); --> how?
+            //properties.name = c_string.unwrap(name).as_ptr();
             properties.vendorID = props.vendor as u32;
             properties.deviceID = props.device as u32;
-            // properties.adapterType = props.device_type; -> warning: enum fields dont match
-            // properties.backendType = props.backend;
+            properties.adapterType = match props.device_type {
+                wgt::DeviceType::Other => native::WGPUAdapterType_Unknown,
+                wgt::DeviceType::IntegratedGpu => native::WGPUAdapterType_IntegratedGPU,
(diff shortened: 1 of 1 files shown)


src/device.rs
```diff
@@ -48,6 +48,26 @@ pub unsafe extern "C" fn wgpuAdapterRequestDevice(
...
+pub unsafe extern "C" fn wgpuAdapterGetProperties(
+    adapter: id::AdapterId,
+    properties: &mut native::WGPUAdapterProperties,
+) {
+    //let maybe_props: Result<wgt::AdapterInfo, wgc::instance::InvalidAdapter> = GLOBAL.adapter_get_info(adapter);
+
+    let maybe_props = gfx_select!(adapter => GLOBAL.adapter_get_info(adapter));
+    match maybe_props {
+        Ok(props) => {
+            //properties.name = std::ffi::CString::new(props.name).unwrap().as_ptr();
```
almarklein, line 61: Are we supposed to create the memory and pass the pointer, or should the user prepare some memory for us to write to?

DevOrc: Either way it might be smart to leave a comment in the header or add an example for this so that is clear who owns what

src/device.rs
```diff
@@ -48,6 +48,26 @@ pub unsafe extern "C" fn wgpuAdapterRequestDevice(
     (callback.unwrap())(device_id, userdata);
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn wgpuAdapterGetProperties(
```
kvark, line 52: We shouldn't add arbitrary entry points to the library, this will break compatibility with webgpu-headers. Instead these extra things should be passed as the extension structures to the existing methods.

almarklein: https://github.com/webgpu-native/webgpu-headers/blob/8b8c2a77dd5a94a64d987b664e2ec320e34a3dc5/webgpu.h#L1074 ;)

kvark: oh well, thanks for the reference :sweat_smile:

almarklein:
It's kinda WIP, but probably good enough for now because I don't have a lot of time right now. It can be made more complete later. Would be good to have a resolution on the ownership of that string though.

kvark:
Is it possible for us to keep an `CString` around for the adapter names?

almarklein:
You know that better than I do :P  I expect it's possible, but I wouldn't know how to do that in Rust.

src/device.rs
almarklein: I think the memory should be created by wgpu-native, because there it is know how large the string is.

almarklein:
I implemented all the fields of `WGPUAdapterProperties` except `name`, so we can also see what GPU and backend have been selected.

For the `name` string, we'd need to globally store the `CString` objects. There are some answers on stackoverflow that explain how to do this, but this seems nontrivial, so I'm going to leave that for another day/person.

kvark (approved):
Thanks!
Gotta tell, I'm not excited about the global thing for the name :/ I hope we'll find a better way.
bors r+

PR #106, opened 2021-05-31
Support for setting TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES

almarklein:
This adds support for specifying native features. Implemented only one feature for now, but is easy to extend later. I called it `nativeFeatures` because it can be expected that `webgpu.h` will specify generic features at some point.


kvark (approved):
bors r+

PR #107, opened 2021-05-31
Fix wgpuQueueWriteTexture signature

### LINKED ISSUE #105: wgpuQueueWriteTexture panic on MacOS
I'm running into rust panics on wgpuQueueWriteTexture on MacOS

thread '<unnamed>' panicked at 'index out of bounds: the len is 1 but the index is 3875379856', /Users/qwerty/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/7ce535c/wgpu-core/src/hub.rs:125:45

Here is a modified examples/triangle/main.c which reproduce this issue:

https://gist.github.com/ryhor-spivak/a119e6caaa15bd69eb91f9bdb1b844e9

I suspect there is a difference in some structure layout between C and Rust code as adding stack variables around WGPUTextureDescriptor / WGPUImageCopyTexture / WGPUTextureDataLayout / WGPUExtent3D changes index value in panic message.

Same code works fine on Windows with Vulkan backend.

almarklein:
closes #105 

@DevOrc like this? Looking at the header:
```h
GPU_EXPORT void wgpuQueueWriteTexture(WGPUQueue queue, WGPUImageCopyTexture const * destination, void const * data, 
    size_t dataSize, WGPUTextureDataLayout const * dataLayout, WGPUExtent3D const * writeSize);
```

I suppose everything `const *` must be either `*const` or `&` in Rust, but how to decide between these?

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -8,10 +8,10 @@ use wgc::{
 #[no_mangle]
 pub unsafe extern "C" fn wgpuCommandEncoderFinish(
     encoder: id::CommandEncoderId,
-    desc: &native::WGPUCommandBufferDescriptor,
+    descriptor: &native::WGPUCommandBufferDescriptor,
 ) -> id::CommandBufferId {
     let desc = wgt::CommandBufferDescriptor {
-        label: OwnedLabel::new(desc.label).into_cow(),
+        label: OwnedLabel::new(descriptor.label).into_cow(),
     };
 
     check_error(gfx_select!(encoder => GLOBAL.command_encoder_finish(encoder, &desc)))
--- a/src/device.rs
+++ b/src/device.rs
@@ -352,11 +352,11 @@ pub unsafe extern "C" fn wgpuQueueWriteBuffer(
 #[no_mangle]
 pub unsafe extern "C" fn wgpuQueueWriteTexture(
     queue: id::QueueId,
-    destination: native::WGPUImageCopyTexture,
+    destination: &native::WGPUImageCopyTexture,
     data: *const u8, // TODO: Check - this might not follow the header
     data_size: usize,
-    data_layout: native::WGPUTextureDataLayout,
-    write_size: native::WGPUExtent3D,
+    data_layout: &native::WGPUTextureDataLayout,
+    write_size: &native::WGPUExtent3D,
 ) {
     let slice = make_slice(data, data_size);
     gfx_select!(queue => GLOBAL.queue_write_texture(
--- a/src/device.rs
+++ b/src/device.rs
@@ -83,17 +83,17 @@ pub unsafe extern "C" fn wgpuDeviceCreateShaderModule(
 #[no_mangle]
 pub extern "C" fn wgpuDeviceCreateBuffer(
     device: id::DeviceId,
-    desc: &native::WGPUBufferDescriptor,
+    descriptor: &native::WGPUBufferDescriptor,
 ) -> id::BufferId {
-    let usage = wgt::BufferUsage::from_bits(desc.usage).expect("Buffer Usage Invalid.");
-    let label = OwnedLabel::new(desc.label);
+    let usage = wgt::BufferUsage::from_bits(descriptor.usage).expect("Buffer Usage Invalid.");
+    let label = OwnedLabel::new(descriptor.label);
     check_error(gfx_select!(device => GLOBAL.device_create_buffer(
         device,
         &wgt::BufferDescriptor {
             label: label.as_cow(),
-            size: desc.size,
+            size: descriptor.size,
             usage,
-            mapped_at_creation: desc.mappedAtCreation,
+            mapped_at_creation: descriptor.mappedAtCreation,
         },
         PhantomData
     )))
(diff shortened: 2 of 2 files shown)


DevOrc:
IIRC functions are supposed to use `*const` / `*mut` because Rust has no way of verifying if the reference is the right type, non-null, etc. However, I have been using references  whenever a reference is assumed to be non-null. This results in undefined behavior if the pointer is invalid. Eventually we should probably use pointer types and then call [ptr.as_ref()](https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ref)

kvark:
@DevOrc I'm not sure it's worth falling back to the raw pointers for these cases. If webgpu-native header says that the pointer must be a valid pointer, then we should just take this assumption and expect a reference. Otherwise, we can't guarantee the safety of the pointer anyway. We can only check for NULL, but there is also a lot of pointer values that we can't check for. It can be garbage for all we know.

kvark (approved):

almarklein:
This issue scared me a bit, so I wrote a quick script to collect the arguments in the C functions that are pointers (from the headerfile, and check that these are also a pointer or ref in our Rust wrappers.

Happy to find that this was the only case. This also shows some of the variations to handle pointers.
```
== Checking command.rs
In wgpuCommandEncoderBeginComputePass:
  ✔ descriptor: &native::WGPUComputePassDescriptor
In wgpuCommandEncoderBeginRenderPass:
  ✔ descriptor: &native::WGPURenderPassDescriptor
In wgpuCommandEncoderCopyBufferToTexture:
  ✔ source: &native::WGPUImageCopyBuffer
  ✔ destination: &native::WGPUImageCopyTexture
  ✔ copysize: &native::WGPUExtent3D
In wgpuCommandEncoderCopyTextureToBuffer:
  ✔ source: &native::WGPUImageCopyTexture
  ✔ destination: &native::WGPUImageCopyBuffer
  ✔ copysize: &native::WGPUExtent3D
In wgpuCommandEncoderCopyTextureToTexture:
  ✔ source: &native::WGPUImageCopyTexture
  ✔ destination: &native::WGPUImageCopyTexture
  ✔ copysize: &native::WGPUExtent3D
In wgpuCommandEncoderFinish:
  ✔ descriptor: &native::WGPUCommandBufferDescriptor
In wgpuComputePassEncoderSetBindGroup:
  ✔ dynamicoffsets: *const u32
In wgpuRenderPassEncoderSetBindGroup:
  ✔ dynamicoffsets: *const u32
In wgpuRenderPassEncoderSetBlendColor:
  ✔ color: &native::WGPUColor
== Checking conv.rs
== Checking device.rs
In wgpuAdapterRequestDevice:
  ✔ descriptor: &native::WGPUDeviceDescriptor
  ✔ userdata: *mut ::std::os::raw::c_void
In wgpuBufferMapAsync:
  ✔ userdata: *mut u8
In wgpuDeviceCreateBindGroup:
  ✔ descriptor: &native::WGPUBindGroupLayoutDescriptor
In wgpuDeviceCreateBindGroupLayout:
  ✔ descriptor: &native::WGPUBindGroupLayoutDescriptor
In wgpuDeviceCreateBuffer:
  ✔ descriptor: &native::WGPUBufferDescriptor
In wgpuDeviceCreateCommandEncoder:
  ✔ descriptor: &native::WGPUCommandEncoderDescriptor
In wgpuDeviceCreateComputePipeline:
  ✔ descriptor: &native::WGPUComputePipelineDescriptor
In wgpuDeviceCreatePipelineLayout:
  ✔ descriptor: &native::WGPUPipelineLayoutDescriptor
In wgpuDeviceCreateRenderPipeline:
  ✔ descriptor: &native::WGPURenderPipelineDescriptor
In wgpuDeviceCreateSampler:
  ✔ descriptor: &native::WGPUSamplerDescriptor
In wgpuDeviceCreateShaderModule:
  ✔ descriptor: &native::WGPUShaderModuleDescriptor
In wgpuDeviceCreateSwapChain:
  ✔ descriptor: &native::WGPUSwapChainDescriptor
In wgpuDeviceCreateTexture:
  ✔ descriptor: &native::WGPUTextureDescriptor
In wgpuInstanceRequestAdapter:
  ✔ options: &native::WGPURequestAdapterOptions
  ✔ userdata: *mut std::os::raw::c_void
In wgpuQueueSubmit:
  ✔ commands: *const id::CommandBufferId
In wgpuQueueWriteBuffer:
  ✔ data: *const u8
In wgpuQueueWriteTexture:
  ✔ destination: &native::WGPUImageCopyTexture
  ✔ data: *const u8
  ✔ datalayout: &native::WGPUTextureDataLayout
  ✔ writesize: &native::WGPUExtent3D
In wgpuTextureCreateView:
  ✔ descriptor: &native::WGPUTextureViewDescriptor
== Checking lib.rs
In wgpuInstanceCreateSurface:
  ✔ descriptor: *const native::WGPUSurfaceDescriptor
In wgpuSurfaceGetPreferredFormat:
  ✔ userdata: *mut std::os::raw::c_void
== Checking logging.rs
0 fails
```

kvark:
Amazing stuff! Would we be able to integrate such a check into CI?
I took the liberty of adjusting the title of the PR. Please use readable titles that don't require redirection :)

kvark (approved):
bors r+

almarklein:
> Please use readable titles that don't require redirection :)

Sorry about that :)

> Would we be able to integrate such a check into CI?

I think we could. It's a Python script though, not sure how you feel about adding that to the repo. It's also a bit quick-n-dirty a.t.m. so it'd need some more love first.

kvark:
bors retry

PR #109, opened 2021-06-02
Enable setting texture dim limits, and limites related to storage buffers

almarklein:
More limits :)

Also made it such that specifying zero will cause the default to be selected. That way the C-struct `WGPUDeviceExtras` can be instantiated without having to specify all limits.

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -202,14 +202,37 @@ pub fn map_device_descriptor<'a>(
         {
             features |= wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES;
         }
+        let mut limits = wgt::Limits::default();
+        if extras.maxTextureDimension1D != 0 {
+            limits.max_texture_dimension_1d = extras.maxTextureDimension1D;
+        }
+        if extras.maxTextureDimension2D != 0 {
+            limits.max_texture_dimension_1d = extras.maxTextureDimension2D;
+        }
+        if extras.maxTextureDimension3D != 0 {
+            limits.max_texture_dimension_2d = extras.maxTextureDimension3D;
+        }
+        if extras.maxTextureArrayLayers != 0 {
+            limits.max_texture_array_layers = extras.maxTextureArrayLayers;
+        }
+        if extras.maxBindGroups != 0 {
+            limits.max_bind_groups = extras.maxBindGroups;
+        }
+        if extras.maxDynamicStorageBuffersPerPipelineLayout != 0 {
+            limits.max_dynamic_storage_buffers_per_pipeline_layout =
+                extras.maxDynamicStorageBuffersPerPipelineLayout;
+        }
+        if extras.maxStorageBuffersPerShaderStage != 0 {
+            limits.max_storage_buffers_per_shader_stage = extras.maxStorageBuffersPerShaderStage;
+        }
+        if extras.maxStorageBufferBindingSize != 0 {
+            limits.max_storage_buffer_binding_size = extras.maxStorageBufferBindingSize;
+        }
         (
             wgt::DeviceDescriptor {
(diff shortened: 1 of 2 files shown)


src/conv.rs
```diff
@@ -207,7 +207,17 @@ pub fn map_device_descriptor<'a>(
                 label: OwnedLabel::new(extras.label).into_cow(),
                 features: features,
                 limits: wgt::Limits {
-                    max_bind_groups: extras.maxBindGroups,
+                    max_texture_dimension_1d: extras.maxTextureDimension1D | 8192,
```
kvark, line 210: this is scary!
Why are we using binary OR for them?

kvark, line 210: instead of hard-coding the base limits, we should take `wgt::Limits::default()` and use its values

kvark (changes requested):

src/conv.rs
almarklein: Good point.

almarklein: If by "we" you mean *me*, I'm just using patterns that I know :)  - is this style frowned upon in Rust? How would I do it instead?

kvark: "we" is a polite term here for "you the author plus us in a virtual term". I.e. if I'm trying to say something is wrong, shifting the responsibility from "you" to "we" makes it much easier for you as the author accept it :)

> is this style frowned upon in Rust? How would I do it instead?


It's not exactly about style. Rather, it's not going to handle cases where `maxTextureDimension1D` is not zero. Say, it's 16384. By `|` you'd produce a value `16384 | 8192` which is bigger than what user requested, so it's unexpected.

There are many ways to approach this, but generally it's hard to be DRY.
The simplest correct way would be:
```
max_texture_dimension_1d: if extras.maxTextureDimension1D != 0 { extras.maxTextureDimension1D } else { defaults.max_texture_dimension_1d },
```

almarklein: Thanks for the explanation. I thought you may have thought I copied that code from elsewhere in the codebase or something. But not being a native English may have contributed to the confusion :)

Oh boy, I meant using a logical or (shortcircuit), not a bitwise or.

kvark: ah, I see. Right, C/C++ has this pattern in places. I didn't see it used in Rust.

kvark: also, not a native speaker, so take my advice with a grain of salt :)

almarklein: But of course a logical or won't work since Rust won't implicitly cast to a bool :)

src/conv.rs
```diff
@@ -202,14 +202,37 @@ pub fn map_device_descriptor<'a>(
...
+        }
         (
             wgt::DeviceDescriptor {
                 label: OwnedLabel::new(extras.label).into_cow(),
                 features: features,
-                limits: wgt::Limits {
-                    max_bind_groups: extras.maxBindGroups,
-                    ..wgt::Limits::default()
-                },
+                limits: limits,
```
kvark, line 235: nit: just `limits,`

src/conv.rs
```diff
@@ -202,14 +202,37 @@ pub fn map_device_descriptor<'a>(
         {
             features |= wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES;
         }
+        let mut limits = wgt::Limits::default();
+        if extras.maxTextureDimension1D > 0 {
```
kvark, line 206: nit: since it's unsigned, maybe`!=` makes it a bit more clear?

kvark (commented):
Thank you!
A few uber nits here, plus please squash it.

src/conv.rs
almarklein: Updated.

kvark (approved):
Great!
bors r+

almarklein:
Nits applied, and squashed.

PR #110, opened 2021-06-03
flatten path to webgpu.h in release

### LINKED ISSUE #108: Release version has incorrect header path
Inside the release zip `wgpu.h` includes `webgpu-headers/webgpu.h`, however the file structure is flattened and it would be correct to include `webgpu.h` instead.

almarklein:
Closes #108


kvark (approved):
bors r+

PR #111, opened 2021-06-04
Fix path flattening in makefile

almarklein:
I should have tested it properly in #110, sorry.

The flattening now happens like this: `sed` is used to replace "webgpu-headers/" with nothing. The resulting changed file is saved as `wgpu.h` (rather than `ffi/wgpu.h`), and this file is included in the zipfile.

I moved away from `sed -i` (inplace) because that API differs between Linux and MacOS:/


kvark (approved):
bors r+

PR #117, opened 2021-07-03
Move to newest wgpu

Sineaggi:
- Adds direct dependency on naga. 
- minor cleanups in examples
- brought naga code over from wgpu-rs
- latest wgpu re-enables surface creation via calayer


Cargo.toml
```diff
@@ -23,23 +23,27 @@ default = []
...
 
 [dependencies]
 lazy_static = "1.1"
 raw-window-handle = "0.3"
 paste = "1.0"
 log = "0.4"
+[dependencies.naga]
+git = "https://github.com/gfx-rs/naga"
+rev = "0b9af95793e319817e74a30601cbcd4bad9bb3e6"
+features = ["wgsl-in", "spv-in"]
```
kvark, line 46: shouldn't need this, since wgpu-native doesn't use `wgsl-in` directly. Only `spv-in` is needed here

kvark (approved):

PR #124, opened 2021-07-30
Update to latest wgpu-core

almarklein:
Pinned to current HEAD commit. Let me know if this is not a good commit to pin to.


kvark (approved):
thank you!
bors r+

PR #126, opened 2021-08-06
Update to latest wgpu-core and naga

almarklein:
Another bump, mostly to include newer Naga with better error reporting and new storage class API. No additional changes required.


kvark (approved):
Thanks!
bors r+

Issue #119, opened 2021-07-09
Missing object destroy functions


helikopterodaktyl:
The current interface only has functions for destroying Buffers, Textures and Query Sets. It doesn't have functions for destroying other GPU objects such as Samplers, Render/Compute Pipelines, Bind Groups, ShaderModules and possibly others.

ryhor-spivak:
Now there are *Drop methods in wgpu.h. What is relationship between those and existing wgpuBufferDestroy, wgpuTextureDestroy, wgpuQuerySetDestroy?

cwfitzgerald:
Destroy is immediate destruction of the resource -- any future uses, including those in already existing BindGroups -- is invalid and will throw an error.

Drop says that the user is done with their handle to the object, but already existing BindGroups etc can keep the object alive.

cwfitzgerald:
Closing this as wgpu-native now has the required functions. webgpu-native can bikeshed in it's own time.

kvark:
If you know you may still have resources referencing this, like stale bind groups, or views, but you want to force the destruction anyway.

PR #128, opened 2021-08-31
Implement drop method for all wgpu objects

Ldash4:
Implements bindings to the drop methods for all WGPU objects. I'm new to webgpu and the surrounding projects, so let me know if something is completely wrong here!

This should solve #119

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -60,4 +60,17 @@ uint32_t wgpuGetVersion(void);
 
 void wgpuRenderPassEncoderSetPushConstants(WGPURenderPassEncoder encoder, WGPUShaderStage stages, uint32_t offset, uint32_t sizeBytes, void* const data);
 
+void wgpuBufferDrop(WGPUBuffer buffer);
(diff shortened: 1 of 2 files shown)


almarklein:
@kvark can we expect these methods to become part of webgpu.h?

Ldash4:
I somehow managed to miss a few declarations in the wgpu header

kvark (approved):
Yes, we should get that to webgpu-headers

kvark:
@Ldash4 would you want to make a PR to webgpu-headers?

Ldash4:
@kvark So essentially the same as [this pr](https://github.com/webgpu-native/webgpu-headers/pull/94), but with `xyzDrop` instead?

kvark:
@Ldash4 there is a critical difference: no "add ref" methods. That PR has "XxxReference", which we don't agree on, at least not yet. As for the naming of the "remove ref" method, I don't have strong feelings. "Drop" works for me, "Release" is not the worst either.

tjpalmer:
Please.

PR #130, opened 2021-09-10
Fix the compute example

lancelet:
- Update `shader.wgsl` from wgpu: https://github.com/gfx-rs/wgpu/blob/master/wgpu/examples/hello-compute/shader.wgsl

### DIFF
--- a/examples/compute/shader.wgsl
+++ b/examples/compute/shader.wgsl
@@ -4,7 +4,7 @@ struct PrimeIndices {
 }; // this is used as both input and output for convenience
 
 [[group(0), binding(0)]]
-var<storage> v_indices: [[access(read_write)]] PrimeIndices;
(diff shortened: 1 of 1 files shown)


almarklein (approved):

kvark (approved):

kvark:
@almarklein we should be detecting this on CI...

almarklein:
Yeah, but that'd mean running the examples on CI, which we cannot do yet, can we?

kvark:
We can! We can do a lot of things.
At the very minimum, we can just run Naga in the build step and parse the source code.
We can actually run the native example on Windows CI, which exposes software implementation (WARP) on DX12.
wgpu has been running GPU tests on Windows CI for a while

almarklein:
#131

PR #133, opened 2021-09-23
Update to latest webgpu.h

almarklein:
The most notable changes are:
* The power_preference can now be set in `request_adapter`.
* The descriptor for `request_adapter` now includes required limits. We supported a few limits via the `WGPUDeviceExtras` struc t- these are now removed.
* New methods `wgpuDeviceGetLimits` and `wgpuAdapterGetLimits`.
* The signature of `WGPURequestAdapterCallback` and `WGPURequestDeviceCallback` have changed.
* Struct field rename: compute_stage -> compute.

Tested by running the examples and by running the wgpu-py test suite.

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -1,7 +1,7 @@
 use crate::{make_slice, map_enum, native, Label, OwnedLabel};
+use naga;
 use std::{borrow::Cow, ffi::CStr, num::NonZeroU32};
 use wgc::{id, pipeline::ShaderModuleSource};
-use naga;
 
 map_enum!(
     map_load_op,
@@ -11,7 +11,6 @@ map_enum!(
     Clear,
     Load
 );
-/*  Temporary explicit map functions while webgpu.h uses Clear instead of Discard.
 map_enum!(
     map_store_op,
     WGPUStoreOp,
@@ -20,14 +19,6 @@ map_enum!(
     Discard,
     Store
 );
-*/
-pub fn map_store_op(value: native::WGPUStoreOp) -> wgc::command::StoreOp {
-    match value {
-        native::WGPUStoreOp_Clear => wgc::command::StoreOp::Discard,
-        native::WGPUStoreOp_Store => wgc::command::StoreOp::Store,
-        x => panic!("Unknown store op: {}", x),
-    }
-}
 map_enum!(
     map_address_mode,
     WGPUAddressMode,
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -192,61 +192,127 @@ pub fn map_adapter_options<'a>(
 }
 
 pub fn map_device_descriptor<'a>(
-    _: &native::WGPUDeviceDescriptor,
+    des: &native::WGPUDeviceDescriptor,
     extras: Option<&native::WGPUDeviceExtras>,
 ) -> (wgt::DeviceDescriptor<Label<'a>>, Option<String>) {
+    let native_limits = unsafe { (*des.requiredLimits).limits };
+    let mut features = wgt::Features::empty();
+    let limits = map_limits(native_limits);
     if let Some(extras) = extras {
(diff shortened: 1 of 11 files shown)


src/device.rs
```diff
@@ -93,6 +101,72 @@ pub unsafe extern "C" fn wgpuAdapterGetProperties(
...
+    return false;
+}
+
+fn write_limits_struct(
+    wgt_limits: wgt::Limits,
+    supported_limits: &mut native::WGPUSupportedLimits,
+) {
+    let mut limits = supported_limits.limits; // This makes a copy - we copy back at the end
+    limits.maxTextureDimension1D = wgt_limits.max_texture_dimension_1d;
+    limits.maxTextureDimension2D = 2; //wgt_limits.max_texture_dimension_2d;
```
kvark, line 136: what happened here?

almarklein: woops, a leftover from a test.

src/device.rs
```diff
@@ -93,6 +101,72 @@ pub unsafe extern "C" fn wgpuAdapterGetProperties(
...
+        _ => panic!("Calling wgpuDeviceGetLimits() on an invalid device."),
+    }
+    return false;
+}
+
+fn write_limits_struct(
+    wgt_limits: wgt::Limits,
+    supported_limits: &mut native::WGPUSupportedLimits,
+) {
+    let mut limits = supported_limits.limits; // This makes a copy - we copy back at the end
```
almarklein, line 134: There is probably a way to write this such that `limits` is a reference instead of a copy, so that the final line of this function is not needed.

almarklein:
I rebased. The new checker detected a mismatch in function argument names, which I fixed too.

src/conv.rs
```diff
@@ -201,61 +192,127 @@ pub fn map_adapter_options<'a>(
...
             },
             None,
-        )
+        );
+    }
+}
+
+pub fn map_limits(limits: native::WGPULimits) -> wgt::Limits {
+    let mut wgt_limits = wgt::Limits::default();
+    if limits.maxTextureDimension1D != 0 {
```
almarklein, line 231: mmm, I guess this should test against `WGPU_LIMIT_U32_UNDEFINED` instead of 0.

src/conv.rs
```diff
@@ -201,61 +192,127 @@ pub fn map_adapter_options<'a>(
 }
 
 pub fn map_device_descriptor<'a>(
-    _: &native::WGPUDeviceDescriptor,
+    des: &native::WGPUDeviceDescriptor,
     extras: Option<&native::WGPUDeviceExtras>,
 ) -> (wgt::DeviceDescriptor<Label<'a>>, Option<String>) {
+    let native_limits = unsafe { (*des.requiredLimits).limits };
```
almarklein, line 198: I guess we want to support a null-pointer here. Should we add this struct to `build.rs` to make it optional?

kvark: Hmm, it doesn't appear to me that it should be optional.

src/conv.rs
almarklein: Although ... I just realized this will potentially break a lot of people's code in a surprising way. I can't help feeling slightly uncomfortable about the `WGPU_LIMIT_U32_UNDEFINED`. For many limits a zero makes no sense, making it a suitable value for "default". @kvark what do you think?

almarklein: I wrote the code, but still feel uncomfortable. I tried to put it in words in https://github.com/webgpu-native/webgpu-headers/issues/111

kvark: Thanks! That makes sense to me

almarklein: I propose that we merge this as-is, and once it the API for specifying undefined is final, we change that for all structs in one PR (possibly a separate release too).

kvark (approved):

PR #136, opened 2021-09-24
Run examples on CI

### LINKED ISSUE #131: Check examples on CI
Serves as a sanity check, but also makes sure that our examples are up-to-date with downstream wgpu-core.

Options:
* Build the examples (we already do this).
* Let Naga parse the wgsl in the build step too.
* Actually run the example (on Windows CI with DX12).

almarklein:
Closes #131.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,27 +99,35 @@ jobs:
             make_command: CMAKE_PREFIX_PATH=$GITHUB_WORKSPACE/glfw-3.3/install make VERBOSE=1
             make_shell: cmd
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v2
         with:
           submodules: 'true'
-      - if: matrix.channel == 'nightly'
-        name: Install latest nightly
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -99,27 +99,35 @@ jobs:
...
       - name: Install dependencies
         run: ${{ matrix.install_deps_command }}
         shell: bash
-      - name: cargo test
+      - name: Run tests
         run: cargo test
-      - if: contains(matrix.build_command, 'clippy')
-        name: cargo clippy
+      - name: Build
         run: ${{ matrix.build_command }}
```
almarklein, line 121: The diff is a bit hard to read, because I added a name field to each step. These two lines were originally this:
```
      - if: contains(matrix.build_command, 'clippy')
        name: cargo clippy
        run: ${{ matrix.build_command }}
```

I suspect this is an error introduced when copy-pasting from another step, because it means that the build step is only run when it contains the word "cllippy".

almarklein:
image

kvark (approved):
Nice!

PR #137, opened 2021-09-24
Update to latest wgpu-core

almarklein:
Based off #133. Will rebase once that is merged.

The trickiest part was to use the new surface API from the swapchain API that we're still exposing. I think this should work?

This compiles, and works for compute examples, but when I run other examples, I get `Unexpected backend Empty`, which is raised in a `gfx_select` macro (e.g. [here](https://github.com/gfx-rs/wgpu/blob/37288a657f022ce508f3b1c92eb6cdbe20f4cca9/wgpu-core/src/lib.rs#L201-L220)) -  I'm a bit stuck on this now. @kvark any idea what could be causing this?

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -42,6 +42,7 @@ pub unsafe extern "C" fn wgpuInstanceRequestAdapter(
             &wgt::RequestAdapterOptions {
                 power_preference,
                 compatible_surface,
+                force_fallback_adapter: false,
             },
             wgc::instance::AdapterInputs::Mask(backend_bits, |_| PhantomData),
         )
--- a/src/device.rs
+++ b/src/device.rs
@@ -1,11 +1,14 @@
 use crate::conv::{map_adapter_options, map_device_descriptor, map_shader_module};
 use crate::{check_error, conv, follow_chain, make_slice, native, OwnedLabel, GLOBAL};
+use lazy_static::lazy_static;
 use std::{
     borrow::Cow,
+    collections::HashMap,
     convert::TryInto,
     marker::PhantomData,
     num::{NonZeroU32, NonZeroU64, NonZeroU8},
     path::Path,
+    sync::Mutex,
 };
 use wgc::{gfx_select, id};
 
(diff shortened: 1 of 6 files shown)


src/device.rs
```diff
@@ -646,36 +648,44 @@ pub extern "C" fn wgpuDeviceCreateSwapChain(
...
 #[no_mangle]
-pub extern "C" fn wgpuSwapChainPresent(swap_chain: id::SwapChainId) {
-    //TODO: Header does not return swap chain status?
-    gfx_select!(swap_chain => GLOBAL.swap_chain_present(swap_chain))
+pub extern "C" fn wgpuSwapChainPresent(swap_chain: id::SurfaceId) {
+    let surface_id = swap_chain;
+    let device_id = surface_id; // TODO get device from surface?
+    gfx_select!(device_id => GLOBAL.surface_present(surface_id))
         .expect("Unable to present swap chain");
 }
```
almarklein, line 690: I'm stuck here. From what I understand, the `gfx_select!` macro wants an argument (before the arrow operator) that is an id that can be associated with the backend in use. The `surface_id` does not work; it panics with "Unexpected backend Empty". I've seen [this code in wgpu-rs](https://github.com/gfx-rs/wgpu/blob/9bc59084926dc66a29bfffdc0a960d7cebbf2e62/wgpu/src/backend/direct.rs#L917-L920) to obtain the device from the surface, but I don't know how to reproduce that here. I'm not even sure that's the way to go TBH :)

@kvark Could you please provide some pointers to help me move forward?

kvark: That's indeed a bit of a problem here, and I'm not too happy about how wgpu-rs solves this. It's able to get away because it can have arbitrary payloads to the objects like `Surface`. Can wgpu-native have a payload to `SurfaceId`? If it can't, we need to figure out how to dispatch this internally, i.e. how to implement `fn surface_present` on wgpu-core side that wouldn't need the generic parameter.

almarklein: I seem to have made it work using a global HashMap and `lazy_static`. My downstream tests and examples work!

kvark (approved):
This is a bit dirty, but it's the best we can do right now (until we find a cleaner way in wgpu-core itself). Thank you for solving this!

PR #146, opened 2021-12-11
Update headers and wgpu to ec1d022

### LINKED ISSUE #127: Errors in shader code show complete source code instead of nice Naga errors
Originally reported in https://github.com/gfx-rs/wgpu/pull/1760#issuecomment-904644388

The problem:
``` 
λ python examples\geometry_cube.py
Failed to parse WGSL code for Some(""): no definition in scope for identifier: 'atomicExchange'
thread '<unnamed>' panicked at 'Parsing(NagaParseError { shader_source: "\n        [[block]]\n        struct Struct_u_stdinfo {\n            cam_transform: mat4x4<f32>;\n            cam_transform_inv: mat4x4<f32>;\n            projection_transform: mat4x4<f32>;\n            projection_transform_inv: mat4x4<f32>;\n            physical_size: vec2<f32>;\n            logical_size: vec2<f32>;\n        };\n\n        [[group(0), binding(0)]]\n        var u_stdinfo: Struct_u_stdinfo;\n        \n\n        [[block]]\n        struct Struct_u_wobject {\n            world_transform: mat4x4<f32>;\n            id: i32;\n        };\n\n        [[group(0), binding(1)]]\n        var u_wobject: Struct_u_wobject;\n        \n\n        [[block]]\n        struct [... etc.]
```

In @kvark's words: In wgpu-rs, between the wgpu-core call and the user error report, there is also this piece happening [...] that's not a part of wgpu-native atm. 

https://github.com/gfx-rs/wgpu/blob/0e65b9323a2e872367dace0685a7b3278d7a9678/wgpu/src/backend/direct.rs#L146-L169

kvark:
Also fixes #127

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,9 +329,8 @@ checksum = "9b919933a397b79c37e33b77bb2aa3dc8eb6e165ad809e58ff75bc7db2e34574"
 
 [[package]]
 name = "glow"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4f04649123493bc2483cbef4daddb45d40bbdae5adb221a63a23efdb0cc99520"
+version = "0.11.1"
+source = "git+https://github.com/grovesNL/glow?rev=5851ca6#5851ca65e6b6f4f2c59683245c07de1843c85452"
 dependencies = [
  "js-sys",
  "slotmap",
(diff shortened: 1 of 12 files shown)


kvark:
@almarklein I believe this is a bigger change. Among other things, it has good error messages, and `[[block]]` is removed from WGSL. So maybe we should kick off a publish build.

almarklein:
This is great! I'm also excited for the better error messages :)  I'll give this a spin locally, and push a tag if all is well.

For reference, the changes to the header file (our public API): https://github.com/webgpu-native/webgpu-headers/compare/4b4dbc3253ba03a1a7b7c42e18cead9dfc1142b8...6a4f8dfced7e221097f87b1a1d4aaa2464cefcbd

From what I can tell, the most notable is the rename `WGPUShaderModuleWGSLDescriptor.source` -> `.code` and the addition of `WGPUPowerPreference_Undefined`. Most other changes in the header relate to things that we have not implemented here yet.

PR #148, opened 2021-12-16
Add build for linux-arm7

almarklein:
Todo:
* [x] Test the binary on a Raspberry Pi.
* [x] We should probably change the name of the build and artefact.

I'm not very familiar with all the different [architectures that Rust can build](https://doc.rust-lang.org/nightly/rustc/platform-support.html), but from what I understood, armv7 is a good target if one wants to support e.g. Raspberry Pi (versions 2/3/4) . This architecture is 32bit, so a better name would be arm32 or maybe armv7. Thoughts? Also, would this binary run on aarch64 machines (which is armv8, IIUC)?

The CD build on my fork: https://github.com/almarklein/wgpu-native/actions/runs/1589649623

Some links that helped me find the right tweaks:
* https://chacin.dev/blog/cross-compiling-rust-for-the-raspberry-pi/
* https://github.com/rust-lang/rust/issues/28924
* https://reposhub.com/rust/embedded/japaric-rust-cross.html

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -128,7 +128,7 @@ jobs:
       CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_LINKER: /usr/bin/arm-linux-gnueabihf-gcc
     steps:
     # Common part (same for nearly each build)
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         submodules: true
     - name: Set WGPU_NATIVE_VERSION
@@ -148,7 +148,7 @@ jobs:
       run: make package
     # Upload (same for each build)
     - name: Upload
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         path: dist
         name: dist
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -117,6 +117,43 @@ jobs:
         name: dist
 
   # -----
+  linux-armv7:
+    # Config
+    name: release - linux-armv7
+    runs-on: ubuntu-latest
+    env:
+      TARGET: armv7-unknown-linux-gnueabihf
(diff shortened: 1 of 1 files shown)


kvark (approved):

almarklein:
Good news and bad news :)

The generated library seems to load fine on a RPi4. Unfortunately it fails with:
```
thread '<unnamed>' panicked at 'index out of bounds: the len is 1 but the index is 1
[...] wgpu-core/src/hub.rs:118:45`
```

The referenced wgpu-core code (at the commit matching the used binary):
https://github.com/gfx-rs/wgpu/blob/c3d906856f11821ab75f7318089704c20922359d/wgpu-core/src/hub.rs#L114-L129

For reference, I first updated the Vulkan drivers with `sudo apt install mesa-vulkan-drivers` to 1.1.97.

almarklein:
Might be worth another try with the wgpu-core-bump.

almarklein:
I rebased, triggered a CD build on my fork, and added the binaries to the v0.11.0.1 release. I'll try again next week.

almarklein:
I rebased, build the binaries, and added them to the v0.14.2.3 release. I then tested on a raspberry pi 4. Unfortunately:
```
OSError: cannot load library '/home/pi/dev/py/wgpu-py/wgpu/resources/libwgpu_native-release.so':
/lib/arm-linux-gnueabihf/libc.so.6: version `GLIBC_2.33' not found
(required by /home/pi/dev/py/wgpu-py/wgpu/resources/libwgpu_native-release.so).
```

Also:
```
pi@pi:~ $ ldd --version
ldd (Debian GLIBC 2.28-10+rpt2+rpi1+deb10u1) 2.28
```

So it looks like the glibc version that the binaries are build against are too new. Does anyone know how to make the compile build against an earlier version of glibc? Would selecting an older Ubuntu host machine work?

almarklein:
Closing, since we have support for 64bit ARM Linux via #326, and the way this PR is stuck seems hard to fix.

PR #152, opened 2021-12-17
Attempt to fix Windows CI

kvark:
Problem - https://github.com/gfx-rs/wgpu-native/pull/151#issuecomment-996743109


kvark:
Found https://github.com/rust-lang/rust/issues/91974

almarklein:
Thanks!

Issue #113, opened 2021-06-15
When will panics become errors that can be handled?


almarklein:
Whenever there is a validation error (e.g. a syntax error in the shader) Rust will panic, and bring down the whole process. This makes it hard to do more interactive things (e.g. in dynamic languages like Python). As I've understood it, the panics are a temporary thing and will eventually be replaced with an error mechanism, so that the calling environment can handle the error. 

Are there ideas about what this will look like and when this will be happening?

kvark:
This is already solved at wgpu-core level. All of the things that can error return `Option<Error>` [on the side](https://github.com/gfx-rs/wgpu/blob/19206064dc7779546e3d9ade9f8093935da5613e/wgpu/src/backend/direct.rs#L693). It's up to the caller to figure out what to do with it. For example, wgpu-rs allows users to [set the uncaptured error handler](https://github.com/gfx-rs/wgpu/blob/19206064dc7779546e3d9ade9f8093935da5613e/wgpu/src/backend/direct.rs#L1299). The default handler does indeed panic.

So I think wgpu-native needs to expose the "on_uncaptured_error" callback, similarly.

sqaxomonophonen:
I've begun work on improving the error handling in wgpu-native here:

https://github.com/sqaxomonophonen/wgpu-native/commit/5ebd50e828b726b58394ee30f0b6a352f0c0c6a0

It's not yet finished, but it implements `wgpuDeviceSetUncapturedErrorCallback()`, and currently improves `wgpuDeviceCreateShaderModule()` and `wgpuSwapChainGetCurrentTextureView()` by not panic'ing when something goes wrong. Instead they call your error callback, and return `NULL`.

@kvark, could you have a look and tell me if it looks right, or if you had something else in mind? I'd be happy to finish the work.

kvark:
That's very nice, thank you @sqaxomonophonen ! It will be convenient to provide feedback on a draft PR if you make one.

kvark:
We don't have push/pop error scopes implemented yet here at wgpu-native level, but @sqaxomonophonen just landed the uncaptured error handler and the device lost handlers in #157.

rajveermalviya:
#264 implemented the push/pop error scopes, and now most the internal errors are handle-able by error scopes.
Some that are categorized as fatal errors, still panic: https://github.com/search?q=repo%3Agfx-rs%2Fwgpu-native+handle_error_fatal&type=code
[...]

Issue #22, opened 2019-06-20
Wrap all extern C functions in catch_unwind


rukai:
Another issue arising from discussion in the rust community discord.

Panicking over FFI boundaries is UB, so we should wrap every extern C function with catch_unwind.

Ideally the rust API could avoid the cost of this, but any solution there would still involve catch_unwind'ing in the extern C functions.

kvark:
Would `panic=abort` resolve the issue here?

cwfitzgerald:
This has been solved by newer rust versions which will automatically catch unwinds at the language boundary unless the C-unwind abi is used.

inner-daemons:
@cwfitzgerald I think what you were referencing here was <https://blog.rust-lang.org/2024/09/05/Rust-1.81.0/#abort-on-uncaught-panics-in-extern-c-functions>. But that automatically aborts on panic being caught, which is REALLY bad for user experience. Should this issue be reopened? Wondering because I was confused about how many panics we had (and I believed it was still undefined behavior, but this isn't much better).

PR #157, opened 2022-01-03
Error handling

sqaxomonophonen:
I've removed `check_error()`, and changed API functions using it to return NULL
instead of panic'ing. There are still things that can panic, but most of it
seems to be trivial assertions, like when passing invalid enum values as such.
A grep for `panic`/`unwrap`/`expect` finds these potential panics.

I've also implemented `wgpuDeviceSetUncapturedErrorCallback()` and
`wgpuDeviceSetDeviceLostCallback()`. `wgpuCommandEncoderFinish()` doesn't
invoke error callbacks yet because I don't have a device handy; I saw a
`SURFACE_TO_DEVICE` HashMap elsewhere, so maybe I should copy this pattern and
have a `COMMAND_ENCODER_TO_DEVICE` map?

I've updated `examples/triangle/main.c` to reflect these changes. I get a
`wgt::SurfaceStatus::Outdated` when resizing the window on Linux/Vulkan, so
I've fixed that too (by making a new swap chain and trying again).

(see also: #113, #22)

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,4 +1,4 @@
-use crate::{check_error, conv, make_slice, native, OwnedLabel, GLOBAL};
+use crate::{conv, make_slice, native, OwnedLabel, GLOBAL};
 use std::{borrow::Cow, num::NonZeroU64};
 use wgc::{
     command::{compute_ffi, render_ffi},
@@ -9,12 +9,19 @@ use wgc::{
 pub unsafe extern "C" fn wgpuCommandEncoderFinish(
     encoder: id::CommandEncoderId,
     descriptor: &native::WGPUCommandBufferDescriptor,
-) -> id::CommandBufferId {
+) -> Option<id::CommandBufferId> {
     let desc = wgt::CommandBufferDescriptor {
         label: OwnedLabel::new(descriptor.label).into_cow(),
     };
 
-    check_error(gfx_select!(encoder => GLOBAL.command_encoder_finish(encoder, &desc)))
+    let (id, error) = gfx_select!(encoder => GLOBAL.command_encoder_finish(encoder, &desc));
+    if let Some(_error) = error {
+        // TODO figure out what device the encoder belongs to and call
+        // handle_device_error()
+        None
+    } else {
+        Some(id)
+    }
 }
 
 #[no_mangle]
--- a/src/command.rs
+++ b/src/command.rs
@@ -15,9 +15,10 @@ pub unsafe extern "C" fn wgpuCommandEncoderFinish(
     };
 
     let (id, error) = gfx_select!(encoder => GLOBAL.command_encoder_finish(encoder, &desc));
-    if let Some(_error) = error {
+    if let Some(error) = error {
         // TODO figure out what device the encoder belongs to and call
         // handle_device_error()
+        log::error!("command_encoder_finish() failed: {:?}", error);
         None
     } else {
         Some(id)
--- a/src/device.rs
+++ b/src/device.rs
@@ -723,7 +723,7 @@ pub extern "C" fn wgpuDeviceCreateSwapChain(
         None
     } else {
         SURFACE_TO_DEVICE.lock().unwrap().insert(surface, device);
-        Some(surface)
+        Some(surface) // swap chain_id == surface_id
     }
 }
 
--- a/src/device.rs
+++ b/src/device.rs
@@ -748,23 +748,25 @@ pub extern "C" fn wgpuSwapChainGetCurrentTextureView(
             handle_device_error(device, &error);
             None
         }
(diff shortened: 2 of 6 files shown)


src/command.rs
```diff
@@ -9,12 +9,19 @@ use wgc::{
...
     descriptor: &native::WGPUCommandBufferDescriptor,
-) -> id::CommandBufferId {
+) -> Option<id::CommandBufferId> {
     let desc = wgt::CommandBufferDescriptor {
         label: OwnedLabel::new(descriptor.label).into_cow(),
     };
 
-    check_error(gfx_select!(encoder => GLOBAL.command_encoder_finish(encoder, &desc)))
+    let (id, error) = gfx_select!(encoder => GLOBAL.command_encoder_finish(encoder, &desc));
+    if let Some(_error) = error {
```
kvark, line 18: should at least try to log this error, e.g. with `log::error!()`

src/device.rs
```diff
@@ -676,26 +719,57 @@ pub extern "C" fn wgpuDeviceCreateSwapChain(
...
-        .expect("Unable to get swap chain texture view")
-        .texture_id
-        .unwrap();
-    let desc = wgc::resource::TextureViewDescriptor::default();
-    Some(gfx_select!(texture => GLOBAL.texture_create_view(texture, &desc, PhantomData)).0)
+    let device = get_device_from_surface(surface_id);
+    match gfx_select!(device => GLOBAL.surface_get_current_texture(surface_id, PhantomData)) {
+        Err(error) => {
+            handle_device_error(device, &error);
+            return None;
```
kvark, line 749: why do we have all those returns? just end with an expression

src/lib.rs
```diff
@@ -1,4 +1,6 @@
-use std::{borrow::Cow, marker::PhantomData, sync::Arc};
+use log;
+use std::ffi::CString;
```
kvark, line 2: please follow the existing style - add inside the `use std::{..}` that's already there

src/lib.rs
```diff
@@ -281,3 +274,89 @@ pub unsafe extern "C" fn wgpuSurfaceGetPreferredFormat(
...
+    userdata: *mut std::os::raw::c_void,
+}
+
+type UncapturedErrorCallback = DeviceCallback<native::WGPUErrorCallback>;
+type DeviceLostCallback = DeviceCallback<native::WGPUDeviceLostCallback>;
+
+unsafe impl<T> Send for DeviceCallback<T> {}
+
+lazy_static::lazy_static! {
+    static ref UNCAPTURED_ERROR_CALLBACKS: Arc<Mutex<HashMap<id::DeviceId, UncapturedErrorCallback>>> = Arc::new(Mutex::new(HashMap::new()));
```
kvark, line 289: First, I don't think `Arc` is needed here.

Secondly, can we merge them into a single structure?
```rust
Mutex<Callbacks>
struct Callbacks {
  uncaptured_errors: HashMap<id::DeviceId, UncapturedErrorCallback>,
  device_lost: HashMap<id::DeviceId, DeviceLostCallback>,
}
```

src/lib.rs
```diff
@@ -281,3 +274,89 @@ pub unsafe extern "C" fn wgpuSurfaceGetPreferredFormat(
...
+#[no_mangle]
+pub unsafe extern "C" fn wgpuDeviceSetDeviceLostCallback(
+    device: id::DeviceId,
+    callback: native::WGPUDeviceLostCallback,
+    userdata: *mut std::os::raw::c_void,
+) {
+    DEVICE_LOST_CALLBACKS.lock().unwrap().insert(
+        device,
+        DeviceLostCallback {
+            callback: callback,
```
kvark, line 317: nit: in Rust, just write this as `callback,`

kvark (commented):
a few nits noted!

src/device.rs
sqaxomonophonen: Apologies, I'm very new to Rust, but when I tried without `return`, I get a `expected enum 'Option', found '()'` compile error. Somehow it doesn't resolve them to `Option<id::TextureViewId>` without a return.

kvark: You'd just leave `None` here without `return` or a semicolon

kvark (approved):

src/lib.rs
sqaxomonophonen: `Mutex<Callbacks>` doesn't compile, but I collected them in a `static ref CALLBACKS: Mutex<Callbacks>`

sqaxomonophonen:
Thanks for the review!

src/lib.rs
kvark: yes, that's what I thought :)

Issue #158, opened 2022-01-06
Unable to link with either Clang or MSVC, missing symbols on Win32


obscurelyme:
```
C:\PROGRA~1\LLVM\bin\CLANG_~1.EXE -fuse-ld=lld-link -nostartfiles -nostdlib -g -Xclang -gcodeview -O0 -D_DEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd -Xlinker /subsystem:console CMakeFiles/HelloWgpu.dir/main.cpp.obj -o HelloWgpu.exe -Xlinker /implib:HelloWgpu.lib -Xlinker /pdb:HelloWgpu.pdb -Xlinker /version:0.0   C:/Users/ME/Projects/wgpu-native/target/debug/wgpu_native.lib  -luserenv.lib  -lws2_32.lib  -lDwmapi.lib  -ldbghelp.lib  -ld3dcompiler.lib  -lD3D12.lib  -lD3D11.lib  -lDXGI.lib  -lsetupapi.lib  -lBcrypt.lib -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 -loldnames && cmd.exe /C "cd /D C:\Users\ME\Projects\hello-wgpu\build && C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -noprofile -executionpolicy Bypass -file C:/Users/ME/vcpkg/scripts/buildsystems/msbuild/applocal.ps1 -targetBinary C:/Users/ME/Projects/hello-wgpu/build/HelloWgpu.exe -installedDir C:/Users/ME/Projects/hello-wgpu/build/vcpkg_installed/x64-windows/debug/bin -OutVariable out""
```

I'm looking to build out a small "Hello, World" using wgpu, but I'm getting missing symbols during the linker step. I've followed the steps in the wiki to pull down and compile this library and I have both a debug and release target built on my system. I can see from the clang linker step that the wgpu_native.lib was added to the list of libraries to link with posted above.

`C:/Users/ME/Projects/wgpu-native/target/debug/wgpu_native.lib`

And my main.cpp that I wish to build is quite simple

```
#include <iostream>
#include <webgpu-headers/webgpu.h>
#include <wgpu.h>

#define WGPU_TARGET_MACOS 1
#define WGPU_TARGET_LINUX_X11 2
#define WGPU_TARGET_WINDOWS 3
[...]

PR #159, opened 2022-01-07
`extern "C"` required for C++ compatibility

sqaxomonophonen:
Without `extern "C"` a simple C++ program like this will fail to link:
```c++
#include "wgpu.h"
main() { wgpuSetLogLevel(WGPULogLevel_Trace); }
```

See also: #158

I've copied the `extern "C"` stuff from the `webgpu.h` header. Although, I don't understand why they include the function pointer typedefs in the `extern "C"` scope. I'm pretty sure it has no effect, since `extern "C"` only prevents C++ name mangling for linker symbols.

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -38,6 +38,10 @@ typedef struct WGPUDeviceExtras {
 
 typedef void (*WGPULogCallback)(WGPULogLevel level, const char *msg);
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 void wgpuDevicePoll(WGPUDevice device, bool force_wait);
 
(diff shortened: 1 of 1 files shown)


ffi/wgpu.h
```diff
@@ -38,6 +38,10 @@ typedef struct WGPUDeviceExtras {
 
 typedef void (*WGPULogCallback)(WGPULogLevel level, const char *msg);
 
+#ifdef __cplusplus
```
kvark, line 41: Why only `__cplusplus`? shouldn't we just do this uncondtionally instead?

kvark: oh, I guess it makes no sense or even forbidden in C

kvark (approved):

Issue #161, opened 2022-01-17
Mismatched types in bindings (windows)


guusw:
Building this on windows generates a lot of mismatched types errors like these for me:
```rs
error[E0308]: mismatched types
   --> src\command.rs:113:44
    |
113 |                 load_op: conv::map_load_op(desc.depthLoadOp),
    |                                            ^^^^^^^^^^^^^^^^ expected `i32`, found `u32`
    |
help: you can convert a `u32` to an `i32` and panic if the converted value doesn't fit
    |
113 |                 load_op: conv::map_load_op(desc.depthLoadOp.try_into().unwrap()),
    |                
```

Removing/changing [src/lib.rs@124](https://github.com/gfx-rs/wgpu-native/blob/636629dcad96a1e450741d61e1d8635cac36785e/src/lib.rs#L124) 
```rs
#[cfg(target_os = "windows")]
pub type EnumConstant = i32;
```
to `u32` fixes this for me, so I'd like to change it.
However I don't know if maybe I'm on some other version of some tool or this might break compilation for anyone else

```
> rustc --version
rustc 1.58.0 (02072b482 2022-01-11)
> cargo version
cargo 1.58.0 (7f08ace4f 2021-11-24)
> rustup default
stable-x86_64-pc-windows-msvc (default)
> bindgen --version
bindgen 0.59.2

```

kvark:
Interesting! I wonder if MSVC vs mingw toolchains are also involved in here.

guusw:
It also happens when I use the gnu toolchain
```
rustup default
stable-x86_64-pc-windows-gnu (default)
```

kvark:
Ok, let's do it then.

rajveermalviya:
seems like, this is still an issue when building with `stable-x86_64-pc-windows-gnu` toolchain.
[here my actions run](https://github.com/rajveermalviya/go-webgpu/runs/4936938559)

```
error[E0308]: mismatched types
   --> src\command.rs:113:44
    |
113 |                 load_op: conv::map_load_op(desc.depthLoadOp),
    |                                            ^^^^^^^^^^^^^^^^ expected `u32`, found `i32`
    |
help: you can convert an `i32` to a `u32` and panic if the converted value doesn't fit
```

kvark:
Can we just add this to our CI, please?

guusw:
I'm working on a PR that uses the respective native enum types from bindings.rs directly, replacing EnumConstant.
[...]

PR #162, opened 2022-01-19
Remove difference in EnumConstant type (windows)

guusw:
Referring to #161 

Creating pull request to check CI since it's green without it on windows.


kvark (approved):

guusw:
After switching toolchains a bunch it seems to have resolved itself. Will update/close PR later.

PR #163, opened 2022-01-20
Use alignment limits from wgpu-core

guusw:
The following values seem to be available right now from wgpu-core but not retreived through the limits API:
- minUniformBufferOffsetAlignment
- minStorageBufferOffsetAlignment

Enabling them here.


guusw:
Actually, all of them seem to have been added. Updated.

kvark (approved):
thank you!

PR #168, opened 2022-02-02
Update wgpu to 08ef577 

caiiiycuk:

PR #171, opened 2022-02-03
support for emscripten build

caiiiycuk:
Added support for emscripten build. It's draft because, I am waiting for completion of other PRs (wgpu, headers).
Also, how I can edit wiki to add emscripten instructions?

--
изображение

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -554,7 +554,7 @@ dependencies = [
 [[package]]
 name = "naga"
 version = "0.8.0"
-source = "git+https://github.com/gfx-rs/naga?rev=81dc674#81dc67402a743b4dc6b2d24c0d306cd18799238b"
+source = "git+https://github.com/gfx-rs/naga?rev=8e2e39e#8e2e39e4d8fa5bbb657c3b170b4f6607d703e284"
 dependencies = [
  "bit-set",
  "bitflags",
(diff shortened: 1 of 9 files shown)


caiiiycuk:
Blocked by: 
* #168 
* https://github.com/webgpu-native/webgpu-headers/pull/146
* https://github.com/gfx-rs/wgpu/pull/2450/files
* +wiki

caiiiycuk:
@kvark can we fork webgpu-headers under gfx-rs to apply emscripten workaound? https://github.com/webgpu-native/webgpu-headers/pull/146

kvark:
We can, yeah, although it would be good to not do that:)
Didn't you also try to make our ID to be 32bit on those systems?

caiiiycuk:
> We can, yeah, although it would be good to not do that:) Didn't you also try to make our ID to be 32bit on those systems?

Yes I tried, and this solution also works. But, I really worried that 65536 is too small for index/epoch? Or you think its ok?


--

Another solution that I have in mind is to doing copy of headers and patch them when make:
```sh
make lib-native-emscripten
```

It will cp webgpu.h into emscripten/webgpu.h  and patch it. What do you think is it acceptable?

kvark:
I think 32-bit indices are acceptable, especially with https://github.com/gfx-rs/wgpu/pull/2462 merged.
It's certainly not ideal, but it should work.

caiiiycuk:
ready

kvark (changes requested):

rajveermalviya:
Closing this as on web/wasm Emscripten provides it's own `webgpu.h` implementation which binds to the WebGPU JS functions directly. It doesn't make sense for `wgpu-native` to "support" emscripten if user wants to target browser webgpu.

Only thing `wgpu-native` can provide in emscripten/wasm context is an implementation of `webgpu.h` with WebGL2 backend. If that's what users want, then please open a new issue stating your usecase.

PR #173, opened 2022-02-20
Add device name

BeastLe9enD:
The wgpuAdapterProperties.name field was not working properly because there was the need to allocate a CString, which needs to be freed in some way. I took the approach to create a lazy_static HashMap which tracks the allocated CString by adapter id and prevents allocating the string multiple times. Of course the CString stays inside the hash map until the application exists, but I think this is the best approach without adding additional functions to the api!

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -1,11 +1,11 @@
 use crate::conv::{map_adapter_options, map_device_descriptor, map_shader_module};
 use crate::{conv, follow_chain, handle_device_error, make_slice, native, OwnedLabel, GLOBAL};
 use lazy_static::lazy_static;
-use std::ffi::CString;
 use std::{
     borrow::Cow,
     collections::HashMap,
     convert::TryInto,
+    ffi::CString,
     marker::PhantomData,
     num::{NonZeroU32, NonZeroU64, NonZeroU8},
     path::Path,
--- a/src/device.rs
+++ b/src/device.rs
@@ -1,6 +1,7 @@
 use crate::conv::{map_adapter_options, map_device_descriptor, map_shader_module};
 use crate::{conv, follow_chain, handle_device_error, make_slice, native, OwnedLabel, GLOBAL};
 use lazy_static::lazy_static;
+use std::ffi::CString;
 use std::{
     borrow::Cow,
     collections::HashMap,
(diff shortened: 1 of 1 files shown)


KitsuneAlex:
I was just about to do the same thing! Nice work :)

src/device.rs
```diff
@@ -1,6 +1,7 @@
 use crate::conv::{map_adapter_options, map_device_descriptor, map_shader_module};
 use crate::{conv, follow_chain, handle_device_error, make_slice, native, OwnedLabel, GLOBAL};
 use lazy_static::lazy_static;
+use std::ffi::CString;
```
kvark, line 4: please follow the existing `use` declaration

src/device.rs
```diff
@@ -85,9 +90,12 @@ pub unsafe extern "C" fn wgpuAdapterGetProperties(
...
     match maybe_props {
         Ok(props) => {
-            //let name = std::ffi::CString::new(props.name);
-            //store_object_globally(name); --> how?
-            //properties.name = c_string.unwrap(name).as_ptr();
+            properties.name = ADAPTER_NAMES
+                .lock()
+                .unwrap()
+                .entry(adapter)
+                .or_insert_with(|| CString::new((&props.name) as &str).unwrap())
```
kvark, line 97: why do we need `(&props.name) as &str` as opposed to just `&props.name`?

kvark (changes requested):
It's quite unfortunate, but will suffice as a temporary solution.
Hopefully, webgpu-headers will find a better answer here.

FWIW, another approach could be to make `AdapterId` effectively a `Box<Something>`, where:
```rust
struct Something {
  raw: id::AdapterId,
  name: CString,
}
```
To clarify, I'm not requesting this code to be changed/rewritten. Just providing this FYI.

src/device.rs
BeastLe9enD: With ```&props.name```, the compiler complains about an unsatisfied trait bound: ```the trait bound `Vec<u8>: From<&std::string::String>` is not satisfied```. With ```(&props.name) as &str```, it works fine.

kvark: `props.name.as_str()` would be even better :)

kvark (approved):

PR #174, opened 2022-03-03
use `forceFallbackAdapter`

rajveermalviya:
use `forceFallbackAdapter` from `WGPURequestAdapterOptions` instead of hardcoding `false`


cwfitzgerald (approved):
Thanks!

PR #175, opened 2022-03-05
Provide a 'fake' implementation of wgpuCreateInstance for better API compatibility.

20kdc:


20kdc:
```
 == Checking lib.rs
In wgpuCreateInstance:
wgpuCreateInstance has missing args: {'descriptor'}
```
Not sure what the mistake is here - is it specifically the argument name, or is there something more to this?

src/lib.rs
```diff
@@ -204,6 +204,15 @@ unsafe impl raw_window_handle::HasRawWindowHandle for PseudoRwh {
     }
 }
 
+#[no_mangle]
+pub extern "C" fn wgpuCreateInstance(
+    _: *const native::WGPUInstanceDescriptor
+) -> native::WGPUInstance {
+    // Rationale: See https://github.com/gfx-rs/wgpu-native/issues/116
+    // Because WGPUInstance is an opaque type this library controls and does not define the contents of, this is safe.
+    8 as native::WGPUInstance
```
kvark, line 213: but.. why 8? :)

kvark (commented):
Not sure what the CI found wrong here. Let's try to match the argument name, I guess?

src/lib.rs
20kdc: I don't know if Rust checks pointer alignments... and I don't want to find out.
So I picked 8. Otherwise, 1 would be the most appropriate choice.
It doesn't really matter in any case - it must be non-zero in case an application tests the return value, that is all.

20kdc:
Is there a way for me to manually run the API check thing locally?

almarklein:
> Not sure what the mistake is here - is it specifically the argument name, or is there something more to this?

The tests check that the Rust API matches the header file.

> Is there a way for me to manually run the API check thing locally?

Te specific test that failed is a self-contained Python file that you can run: `checks.py`. The other tests are run with `cargo test`and [some more](https://github.com/gfx-rs/wgpu-native/blob/master/.github/workflows/ci.yml#L133)

kvark (approved):

PR #184, opened 2022-04-06
Update to latest wgpu-core

almarklein:
I get the following error when I try to run any example:
```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`: ShaderModule[0] is no longer alive', /Users/almar/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/988990c/wgpu-core/src/hub.rs:169:9
```

Does anyone have an idea why this is?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "ash"
-version = "0.33.3+1.2.191"
+version = "0.37.0+1.3.209"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc4f1d82f164f838ae413296d1131aa6fa79b917d25bebaa7033d25620c09219"
+checksum = "006ca68e0f2b03f22d6fa9f2860f85aed430d257fec20f8879b2145e7c7ae1a6"
 dependencies = [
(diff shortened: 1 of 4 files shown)


almarklein:
@almarklein Thanks, that seems to be a needed change indeed, but it does not fix the error. I get the same error if I add gibberish in the shader, so it looks like it never gets to parsing/compiling it.

almarklein:
Correction, I do get some examples to run by changing both the `[[xxx]]` decorators, and also using `,`instead of `;` in struct separators.

So, I guess we have a new bug, where syntax errors in the wgsl result in a seemingly unrelated error.

rajveermalviya:
> So, I guess we have a new bug, where syntax errors in the wgsl result in a seemingly unrelated error.

it does report those errors via `UncapturedErrorCallback`.

almarklein:
> it does report those errors via UncapturedErrorCallback.

Aha! Thanks, I did see that come by some time ago. Time to implement that in wgpu-py.

almarklein:
After making the necessary updates in wgpu-py, it's test suite passes with this new version. Even fixes some tests that failed specifically on MacOS.

PR #185, opened 2022-04-10
Fixed some build warnings in the examples.

b0bh00d:
Simple fixes in the example C code to reduce the warning noise.

### DIFF
--- a/examples/capture/main.c
+++ b/examples/capture/main.c
@@ -26,7 +26,7 @@ int main(int argc, char *argv[]) {
                   .chain =
                       (WGPUChainedStruct){
                           .next = NULL,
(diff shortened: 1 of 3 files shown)


examples/capture/main.c
```diff
@@ -39,6 +39,11 @@ int main(int argc, char *argv[]) {
                           .maxBindGroups = 1,
                       },
               },
+          .defaultQueue =
```
kvark, line 42: Interesting. Is this causing a warning?

kvark (commented):
Wow, Bob is here!

kvark (approved):
Ok, I think missing initializers is a feature, not a bug. The spec guarantees them to be zero.
But we can change that warning later.

Issue #186, opened 2022-04-17
Memory statistics logging


sqaxomonophonen:
Hi! I'm using wgpu-native semi-seriously (https://github.com/sqaxomonophonen/gpudl) and I've started worrying about whether or not my programs are leaking resources. So I'm running with a patched wgpu-native:

```diff
diff --git a/src/device.rs b/src/device.rs
index 2ab91fd..b773e51 100644
--- a/src/device.rs
+++ b/src/device.rs
@@ -782,6 +782,7 @@ pub extern "C" fn wgpuSwapChainPresent(swap_chain: id::SurfaceId) {
     let device_id = get_device_from_surface(surface_id);
     gfx_select!(device_id => GLOBAL.surface_present(surface_id))
         .expect("Unable to present swap chain");
+    log::debug!("Memory report: {:?}", GLOBAL.generate_report());
 }
 
 #[no_mangle]
```

E.g. if I forget to `wgpuTextureViewDrop()` the view returned by `wgpuSwapChainGetCurrentTextureView()` I can see the `num_occupied` for `texture_views` go up.

I don't know if it's something you're interested in having in wgpu-native? On one hand, a one-liner patch solves _my_ problem, but on the other hand it sounds like the perfect "bikeshed discussion" :-)
 - There are a dozen ways these kinds of statistics could be exposed, and I clearly chose the easiest, most unintrusive one, picking a function I know I'm going to call roughly once per frame (or more for 2+ windows), and logging from there.
 - Perhaps it ought to be part of `webgpu.h`, that is, after the team resolves this: https://github.com/webgpu-native/webgpu-headers/issues/9
 - Log output could be prettier, but then we're probably talking about a `Display` trait for `GlobalReport` in `wgpu`, which seems overkill?
 - I could use `log::trace!()` instead, but that output is already quite verbose
 - I don't recall ever worrying about leaking resources in OpenGL programs, so maybe I'm partly motivated by the "alpha-software state of everything WebGPU". But maybe OpenGL (~2) also tends to have slightly less resource management?

kvark:
Let's do it, but perhaps in Debug-only?

sqaxomonophonen:
Cool! There's a PR (#187)

PR #187, opened 2022-04-18
Show memory report in debug builds

sqaxomonophonen:
`log::debug!()` after `wgpuSwapChainPresent()` which is assumed to be called roughly once per frame. This is useful for detecting memory leaks.

See also: #186

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -70,6 +70,9 @@ void wgpuSetLogLevel(WGPULogLevel level);
 
 uint32_t wgpuGetVersion(void);
 
+// Returns resource usage C string; caller owns the string and must free() it
+char* wgpuGetResourceUsageString();
+
 void wgpuRenderPassEncoderSetPushConstants(WGPURenderPassEncoder encoder, WGPUShaderStageFlags stages, uint32_t offset, uint32_t sizeBytes, void* const data);
 
 void wgpuBufferDrop(WGPUBuffer buffer);
--- a/src/device.rs
+++ b/src/device.rs
@@ -782,6 +782,8 @@ pub extern "C" fn wgpuSwapChainPresent(swap_chain: id::SurfaceId) {
     let device_id = get_device_from_surface(surface_id);
     gfx_select!(device_id => GLOBAL.surface_present(surface_id))
         .expect("Unable to present swap chain");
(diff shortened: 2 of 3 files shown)


kvark:
I suggested doing it in debug only, and now I'm having second thoughts. Sorry about this!
This memory dump can be quite verbose if done every frame.
Perhaps, we should just expose it as a C function and let the users call it when needed?

sqaxomonophonen:
It could be argued that the WebGPU specification *itself* is lacking a way to get a resource usage report (for the same reason it has `.destroy()` methods; it implies the possibility of resource management problems even in pure JS+WebGPU applications). So any solution we come with is likely going to be temporary? I'm also fine with adding something like this to `wgpu.h`:
```c
void wgpuLogResourceUsage(); // calls your WGPULogCallback with WGPULogLevel=0 regardless of log level
// *OR*:
char* wgpuGetResourceUsageString(); // caller "owns" the returned C-string and must free() it
```
Compared to the implicit logging after `wgpuSwapChainPresent()` it has a higher risk of breaking forward compatibility, but it's easy to fix (compared with all the other forward compatibility issues). It doesn't add unavoidable verbosity, and it also works better with multi-window programs, or programs that don't want a report every frame.

Anything more complicated than this (like getting the report in a C struct), and I think the effort will we wasted here and better spent discussing it with the WebGPU people. Thankfully, running a custom build of `wgpu-native` is easy (:+1:), and I'm fine with it for the time being, so feel free to `.destroy()` this PR if neither solution sits well with you :smiley:

kvark:
`wgpuGetResourceUsageString` sounds good to land here

sqaxomonophonen:
Does this look good?

I tested with the `triangle` example, and this works as expected:
```diff
diff --git a/examples/triangle/main.c b/examples/triangle/main.c
index 0de6675..4c2a8b0 100644
--- a/examples/triangle/main.c
+++ b/examples/triangle/main.c
@@ -342,6 +342,10 @@ int main() {
     wgpuSwapChainPresent(swapChain);
 
     glfwPollEvents();
+
+    char* usage = wgpuGetResourceUsageString();
+    printf("ru: %s\n", usage);
+    free(usage);
   }
 
   glfwDestroyWindow(window);
```

kvark (approved):

PR #191, opened 2022-04-26
Implement push constants

MineGame159:
Implemented push constants and the necessary additions. I have tested it on my Windows machine using the Vulkan backend and it works as expected.

I am new at Rust so hopefully the code is alright.

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -1,8 +1,7 @@
 use crate::{make_slice, map_enum, native, Label, OwnedLabel, follow_chain};
 use naga;
-use std::{borrow::Cow, ffi::CStr, num::NonZeroU32};
+use std::{borrow::Cow, ffi::CStr, num::NonZeroU32, slice};
 use wgc::{id, pipeline::ShaderModuleSource};
-use wgt::{PushConstantRange};
 
 map_enum!(
     map_load_op,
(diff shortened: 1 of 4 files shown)


src/conv.rs
```diff
@@ -1,7 +1,8 @@
-use crate::{make_slice, map_enum, native, Label, OwnedLabel};
+use crate::{make_slice, map_enum, native, Label, OwnedLabel, follow_chain};
 use naga;
 use std::{borrow::Cow, ffi::CStr, num::NonZeroU32};
 use wgc::{id, pipeline::ShaderModuleSource};
+use wgt::{PushConstantRange};
```
kvark, line 5: let's not include this. See how all the code here just refers to the fully qualified type. So this would be `wgt::PushconstantRange`

src/conv.rs
```diff
@@ -234,7 +244,40 @@ pub fn map_device_descriptor<'a>(
...
 
-pub fn map_limits(limits: native::WGPULimits) -> wgt::Limits {
+pub fn map_pipeline_layout_descriptor<'a>(
+    des: &native::WGPUPipelineLayoutDescriptor,
+    extras: Option<&native::WGPUPipelineLayoutExtras>,
+) -> wgc::binding_model::PipelineLayoutDescriptor<'a> {
+    let mut push_constant_ranges: Vec<PushConstantRange> = Vec::new();
+
+    if let Some(extras) = extras {
+        for i in 0..extras.pushConstantRangeCount {
```
kvark, line 254: we should just construct the slice here, i.e. `let raw_push_constant_ranges = slice::from_raw_parts(...);`

src/conv.rs
```diff
@@ -234,7 +244,40 @@ pub fn map_device_descriptor<'a>(
...
+            let range = unsafe { *extras.pushConstantRanges.offset(i as isize) as native::WGPUPushConstantRange };
+            push_constant_ranges.push(wgt::PushConstantRange {
+                stages: wgt::ShaderStages::from_bits(range.stages).expect("Invalid shader stage"),
+                range: core::ops::Range {
```
kvark, line 258: let's do `range: range.start..range.end` instead

kvark (commented):
Great contribution here!
Just a few stylistic notes. Overall looks great to me 👍

kvark (approved):

PR #194, opened 2022-05-20
Implement some functions, mainly debug ones

MineGame159:
Implemented:
- `insert debug marker`, `push debug group`, `pop debug group` functions
- `get bind group layout` functions
- empty implementation of `device destroy` function

WebGPU doesn't have the color parameter for debug markers and groups but WGPU does (even tho it seems like colors are not implemented currently) so I added new functions for them in `wgpu.h`. Hopefully that's fine.

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -74,6 +74,11 @@ uint32_t wgpuGetVersion(void);
 char* wgpuGetResourceUsageString();
 
 void wgpuRenderPassEncoderSetPushConstants(WGPURenderPassEncoder encoder, WGPUShaderStageFlags stages, uint32_t offset, uint32_t sizeBytes, void* const data);
+void wgpuRenderPassEncoderInsertDebugMarkerColor(WGPURenderPassEncoder renderPassEncoder, char const * markerLabel, uint32_t color);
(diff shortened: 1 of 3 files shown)


kvark:
Thank you for the PR!
> WebGPU doesn't have the color parameter for debug markers and groups but WGPU does (even tho it seems like colors are not implemented currently) so I added new functions for them in wgpu.h. Hopefully that's fine.

I don't think it's worth exposing those functions right now. Just pass the color as 0 from the functions agreed by WebGPU.
If color is useful, we should discuss it in webgpu-headers and change the function signatures together with Google.

kvark (approved):

kvark:
CI detected something off

MineGame159:
I don't really understand what it is saying. Do the parameter names have to be exactly the same as in the webgpu headers?

MineGame159:
Should I rename the parameters to match the header?

almarklein:
> Should I rename the parameters to match the header?

Yes, they should be the same as in the header file. That integrity check makes sure that all the args match, but it is not smart enough that with "label" you meant "groupLabel" or "markerLabel" :)

MineGame159:
Sorry it took so long but should be fixed.

MineGame159:
Is there something else I should change?

cwfitzgerald:
Nope! Thank you!

PR #196, opened 2022-06-19
update wgpu

rajveermalviya:
- update webgpu-headers to 402ee24
- update wgpu-core & wgpu-types to f27a978
- `cargo fmt`
- `wgpuComputePassEncoderDispatch` -> `wgpuComputePassEncoderDispatchWorkgroups`
- `wgpuComputePassEncoderDispatchIndirect` -> `wgpuComputePassEncoderDispatchWorkgroupsIndirect`

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -1,4 +1,6 @@
-use crate::conv::{map_adapter_options, map_device_descriptor, map_shader_module, map_pipeline_layout_descriptor};
+use crate::conv::{
+    map_adapter_options, map_device_descriptor, map_pipeline_layout_descriptor, map_shader_module,
+};
 use crate::{conv, follow_chain, handle_device_error, make_slice, native, OwnedLabel, GLOBAL};
 use lazy_static::lazy_static;
 use std::{
@@ -482,7 +484,7 @@ pub unsafe extern "C" fn wgpuQueueSubmit(
 ) {
     let command_buffer_ids = make_slice(commands, command_count as usize);
     gfx_select!(queue => GLOBAL.queue_submit(queue, command_buffer_ids))
-        .expect("Unable to submit queue")
+        .expect("Unable to submit queue");
 }
 
 #[no_mangle]
@@ -535,8 +537,10 @@ pub unsafe extern "C" fn wgpuBufferMapAsync(
             x => panic!("Unknown map mode: {}", x),
         },
         // TODO: Change wgpu-core to follow new API
-        callback: std::mem::transmute(callback.expect("Callback cannot be null")),
-        user_data,
+        callback: wgc::resource::BufferMapCallback::from_c(wgc::resource::BufferMapCallbackC {
+            callback: std::mem::transmute(callback.expect("Callback cannot be null")),
+            user_data,
+        }),
     };
 
     gfx_select!(buffer => GLOBAL.buffer_map_async(buffer, offset as u64 .. (offset + size) as u64, operation))
@@ -545,7 +549,13 @@ pub unsafe extern "C" fn wgpuBufferMapAsync(
 
 #[no_mangle]
 pub unsafe extern "C" fn wgpuDevicePoll(device: id::DeviceId, force_wait: bool) {
-    gfx_select!(device => GLOBAL.device_poll(device, force_wait)).expect("Unable to poll device")
+    gfx_select!(device => GLOBAL.device_poll(
+        device,
+        match force_wait {
+        true => wgt::Maintain::Wait,
+        false => wgt::Maintain::Poll,
+    }))
+    .expect("Unable to poll device");
 }
 
 #[no_mangle]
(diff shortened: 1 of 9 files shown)


rajveermalviya:
cc @almarklein, for changes in `checks.py`

src/lib.rs
```diff
@@ -311,9 +311,13 @@ pub unsafe extern "C" fn wgpuSurfaceGetPreferredFormat(
...
 ) -> native::WGPUTextureFormat {
-    let preferred_format = match wgc::gfx_select!(adapter => GLOBAL.surface_get_preferred_format(surface, adapter))
+    let preferred_format = match wgc::gfx_select!(adapter => GLOBAL.surface_get_supported_formats(surface, adapter))
     {
-        Ok(format) => conv::to_native_texture_format(format),
+        Ok(formats) => conv::to_native_texture_format(
+            *formats
+                .first()
+                .expect("Could not get preferred swap chain format"),
+        ),
```
almarklein, line 320: For reference: https://github.com/gfx-rs/wgpu/blob/0d4d3f4ceb0c07de38064533c3409a2f8c1d8d95/wgpu/src/lib.rs#L3448-L3454  the first format is the preferred one.

almarklein (approved):
LGTM! Thanks!

rajveermalviya:
@cwfitzgerald PTAL

src/device.rs
```diff
@@ -535,8 +537,10 @@ pub unsafe extern "C" fn wgpuBufferMapAsync(
             x => panic!("Unknown map mode: {}", x),
         },
         // TODO: Change wgpu-core to follow new API
```
cwfitzgerald, line 539: This still relevant?

src/device.rs
```diff
@@ -545,7 +549,13 @@ pub unsafe extern "C" fn wgpuBufferMapAsync(
 
 #[no_mangle]
 pub unsafe extern "C" fn wgpuDevicePoll(device: id::DeviceId, force_wait: bool) {
```
cwfitzgerald, line 551: We should change the signature of this to allow for waiting for a particular submission

src/lib.rs
```diff
@@ -311,9 +311,13 @@ pub unsafe extern "C" fn wgpuSurfaceGetPreferredFormat(
     surface: id::SurfaceId,
     adapter: id::AdapterId,
 ) -> native::WGPUTextureFormat {
```
cwfitzgerald, line 313: Keeping this function is fine - to keep compliant with webgpu.h - but we should also have a function which returns all the formats so the user can choose.

cwfitzgerald (changes requested):
Looks good, some comments

src/device.rs
cwfitzgerald: Oh yeah, we should add a wgpuQueueSubmitForIndex or something that returns a submission index

src/device.rs
rajveermalviya: removed

src/device.rs
rajveermalviya: done

src/lib.rs
rajveermalviya: added `wgpuSurfaceGetSupportedFormats`

src/lib.rs
```diff
@@ -311,14 +311,43 @@ pub unsafe extern "C" fn wgpuSurfaceGetPreferredFormat(
...
     };
     return preferred_format;
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn wgpuSurfaceGetSupportedFormats(
+    surface: id::SurfaceId,
+    adapter: id::AdapterId,
+    count: Option<&mut usize>,
+) -> *const native::WGPUTextureFormat {
```
cwfitzgerald, line 331: We can't actually return an owned allocation here because if we are running in a DLL (on windows) we actually have a different heap than the main process so them calling free would be invalid.

We have a couple options here:
- Force the user to allocate the memory and having them pass in a pointer and a length. The question becomes how to communicate to the user how much data to allocate. We could suggest "give us space for 8 formats" which is likely always enough. Otherwise we would need a function call to give them a length.
- Create a wgpuSurfaceSupportedFormatDelete function where we take back the pointer and delete it in our DLL. We'd need to be careful about this though (for example, for vectors length must == capacity when converting to a single length and pointer.

cwfitzgerald (changes requested):
Much closer, one more comment.

src/lib.rs
almarklein: Another option that we have used elsewhere in this repo, IIRC:
* Maintain ownership but never delete it. I.e. keep it a global store.

rajveermalviya: added `wgpuFree`; it can be used to free allocations from rust side.

rajveermalviya: well msvc doesn't like this macro
```c
#define WGPU_FREE(ptr, count) wgpuFree(ptr, count * sizeof(*ptr), _Alignof(typeof(*ptr)))

// usage
WGPU_FREE(usage, 1);
```

because it doesn't support expressions in `_Alignof` & also doesn't support `typeof` operator.
so macro will have to take in the type too, like
```c
#define WGPU_FREE(type, ptr, count) wgpuFree(ptr, count * sizeof(type), _Alignof(type))

// usage
WGPU_FREE(char, usage, 1); // users will have to be careful not to pass the pointer of type.
```

rajveermalviya: and here's how it looks without macro

```c
char *usage = wgpuGetResourceUsageString();
printf("%s\n", usage);
wgpuFree(usage, 1 * sizeof(char), _Alignof(char));
```

personally i would just avoid the macro

cwfitzgerald: Cool, sounds good to me.

cwfitzgerald (changes requested):
One very minor nit, then we're good to go

ffi/wgpu.h
```diff
@@ -3,6 +3,8 @@
 
 #include "webgpu-headers/webgpu.h"
 
+#define WGPU_FREE(type, ptr, count) wgpuFree(ptr, count * sizeof(type), _Alignof(type))
```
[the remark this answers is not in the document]
rajveermalviya: added doc

cwfitzgerald (approved):
Thanks!

PR #198, opened 2022-06-21
Implement VERTEX_WRITABLE_STORAGE native feature

MineGame159:
Support for enabling `VERTEX_WRITABLE_STORAGE` native feature.
This bumps the enum size from 32 bit to 64 bit but I don't think there is any way around this.

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -14,7 +14,8 @@ typedef enum WGPUNativeSType {
 
 typedef enum WGPUNativeFeature {
     WGPUNativeFeature_PUSH_CONSTANTS = 0x04000000,
-    WGPUNativeFeature_TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES = 0x10000000
+    WGPUNativeFeature_TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES = 0x10000000,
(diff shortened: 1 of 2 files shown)


MineGame159:
The Windows builds just hate me here. It is failing because of warnings which are not a problem on other builds.

MineGame159:
Ok well update to latest commit and there is a problem because the WGPUNativeFeature type is now 64 bit so `wgpu****EnumerateFeatures` doesn't work for `VERTEX_WRITABLE_STORAGE` but `wgpu****HasFeature` somehow seems to work.

But there is still the same issue of Windows builds failing because of warnings from other code.

rajveermalviya:
I think we should change how we take in native-only features,
instead of having a separate [`WGPUDeviceExtras.nativeFeatures`](https://github.com/gfx-rs/wgpu-native/blob/2d32686a4a8f17f6b90926dc770b0480a28db942/ffi/wgpu.h#L49) bitflags field.
we should take it in via [`WGPUDeviceDescriptor.requiredFeatures`](https://github.com/webgpu-native/webgpu-headers/blob/402ee247cfbd9a65673e0f9141d80ccf38e79959/webgpu.h#L1133) slice along with other features.

will send a PR in a bit to address that.

MineGame159:
Ok yeah, that makes sense.

MineGame159:
Merged into #209.

PR #200, opened 2022-06-23
cd: use clang-5.0 from `stretch-backports` repo

rajveermalviya:


cwfitzgerald (approved):
NGL, I only just realized "cd" in the title stood for the later part of "CI/CD" - smh.

PR #201, opened 2022-06-24
Implement `wgpuGenerateReport`

rajveermalviya:
replaces `wgpuGetResourceUsageString`

string returned by `wgpuGetResourceUsageString` had to be parsed by wrapper into a language native struct.
populating a struct & accessing the fields directly is much better.


cwfitzgerald (approved):
Really great stuff!

PR #203, opened 2022-07-03
update wgpu-core to 0.13

rajveermalviya:
- wgpu-core, wgpu-types to 0.13
- naga to 0.9
- fix for sparse RenderPassColorAttachment
- fix for sparse ColorTargetState
- add more TextureFormat mappings
- fix conv::map_feature
- add `wgpuSurfaceGetSupportedPresentModes`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -68,30 +68,74 @@ fn main() {
             .raw_line(format!("type {} = wgc::id::{};", old_name, new_name));
     }
 
-    // WGPUBindGroupEntry has fields that are id's which can be 0
+    // WGPUBindGroupEntry.{buffer, sampler, textureView} is nullable
     builder = builder.blocklist_item("WGPUBindGroupEntry").raw_line(
         "#[repr(C)]
             pub struct WGPUBindGroupEntry {
-                pub nextInChain: * const crate::native::WGPUChainedStruct,
+                pub nextInChain: *const WGPUChainedStruct,
                 pub binding: u32,
-                pub buffer: Option<wgc::id::BufferId>,
+                pub buffer: Option<WGPUBuffer>,
                 pub offset: u64,
                 pub size: u64,
-                pub sampler: Option<wgc::id::SamplerId>,
-                pub textureView: Option<wgc::id::TextureViewId>,
+                pub sampler: Option<WGPUSampler>,
+                pub textureView: Option<WGPUTextureView>,
             }",
     );
 
-    // WGPURequestAdapterOptions.compatibleSurface can be Null
+    // WGPURequestAdapterOptions.compatibleSurface is nullable
     builder = builder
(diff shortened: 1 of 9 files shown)


cwfitzgerald:
If we continue to target git, which I think wgpu-native should, the patches are already merged, just not yet backport Ed.

rajveermalviya:
CI failure is unrelated, rust nightly toolchain seems borked on windows.

rajveermalviya:
> continue to target git, which I think wgpu-native should

updated

build.rs
```diff
@@ -68,33 +68,77 @@ fn main() {
             .raw_line(format!("type {} = wgc::id::{};", old_name, new_name));
     }
 
-    // WGPUBindGroupEntry has fields that are id's which can be 0
+    // WGPUBindGroupEntry.{buffer, sampler, textureView} is nullable
```
cwfitzgerald, line 71: I'm not going to veto this now - but this seems kinda unsustainable long term - why do we need to redefine these structs completely?

ffi/wgpu.h
```diff
@@ -55,8 +56,15 @@ typedef struct WGPUDeviceExtras {
...
     WGPUChainedStruct chain;
     uint32_t maxPushConstantSize;
+    uint64_t maxBufferSize;
 } WGPURequiredLimitsExtras;
 
+typedef struct WGPUSupportedLimitsExtras {
+    WGPUChainedStructOut chain;
+    uint32_t maxPushConstantSize;
+    uint64_t maxBufferSize;
+} WGPUSupportedLimitsExtras;
```
cwfitzgerald, line 66: Why do we need both?

cwfitzgerald (commented):
Looks good, one quick comment before we merge; thanks for doing this!

build.rs
rajveermalviya: So, i looked into bindgen if it allows making struct fields optional instead of rewriting whole struct but was unsuccessful to find it.

Other thing we could do is instead of [generating](https://github.com/gfx-rs/wgpu-native/blob/91e14b8ffda0a5ec1d87fd27b7b9e8e62495720d/build.rs#L68) `WGPU[Type]` -> `[Type]Id` generate `WGPU[Type]` -> `Option<[Type]Id>` (which I think is what should have been done earlier) and do `unwrap`/`handle_error`s in each function calls, though don't know what the overhead of it would be. Then completely remove all struct rewrites.

cwfitzgerald (approved):

build.rs
rajveermalviya: ack

cwfitzgerald (approved):
Great Stuff!

rajveermalviya:
> Great Stuff!

spoke too soon lol, but should be ready to merge now

cwfitzgerald:
Big whoops

Issue #212, opened 2022-08-13
Report errors in a C-usable format


spindlebink:
As far as I can tell, there's no way to modify the format error messages are sent in, so WGPU errors get sent to their handler as a single line of unformatted object notation. It goes without saying that this makes debugging highly annoying.

For the Rust side of WGPU, this likely isn't a problem, since Rust has its own error handling and doesn't have to rely on a single C-side callback. But on the `wgpu-native` side, it'd be nice if errors could be reported in a format C can use--JSON at the very least would be usable.

rajveermalviya:
error handling definitely needs an improvement, I am fine with serializing errors to json (but not sure how that would impact binary size).
ideal solution would be to list enums of every possible errors and return them for `type` arg in the callback, but even then you will have to parse the message for meaningful conversion to other language structs.

otoh you could [parse the current error message](https://github.com/pygfx/wgpu-py/blob/aa382b97e70346e4438bf0b787797eafa05c1204/wgpu/backends/rs_helpers.py#L240-L306) ^^

rajveermalviya:
>Related to this issue: the WGPU log callback does not currently provide a way to provide a userdata pointer to its call, unlike the other callbacks, and doing so is pretty important for interoperability between non-Rust/non-C languages and the API.

Opened #213 for this

PR #213, opened 2022-08-14
support userdata in log callback

rajveermalviya:
https://github.com/gfx-rs/wgpu-native/issues/212#issue-1337939897

>Related to this issue: the WGPU log callback does not currently provide a way to provide a userdata pointer to its call, unlike the other callbacks, and doing so is pretty important for interoperability between non-Rust/non-C languages and the API.


cwfitzgerald (approved):
Nice

Issue #214, opened 2022-09-09
Inconsistency in undefined values in requiredLimits


guusw:
emscripten treats WGPU_LIMIT_U32_UNDEFINED/WGPU_LIMIT_U64_UNDEFINED in limits as them being unset. ([reference](https://github.com/emscripten-core/emscripten/blob/829b9337f709fc8055170ed3c2c9653ad6ee970a/src/library_webgpu.js#L2549))
dawn does the same ([reference](https://dawn.googlesource.com/dawn/+/refs/heads/main/src/dawn/native/Limits.cpp#136))
wgpu-native treats 0 instead as them being unset.

I would've prefered 0 instead but it seems like 2 implementations already use WGPU_LIMIT_UXX_UNDEFINED instead so it might be good to also follow so switching implementations becomes more seamless.

PR #218, opened 2022-10-11
check for undefined constants instead of zero values

rajveermalviya:
Align better with other webgpu-headers implementations for unset/undefined values.
Currently we take zero for indicating unset/undefined values. But webgpu-headers defines [constants](https://github.com/webgpu-native/webgpu-headers/blob/69c87bd5017fefd058fdd2b997831da8933fc9c7/webgpu.h#L55-L61) which should be used instead.

This introduces one breaking change; `WGPULimits` struct will have to be declared like this:

```diff
- (WGPULimits){
-     .maxBindGroups = 1,
- }
+ (WGPULimits){
+     .maxBindGroups = 1,
+     .maxTextureDimension1D = WGPU_LIMIT_U32_UNDEFINED,
+     .maxTextureDimension2D = WGPU_LIMIT_U32_UNDEFINED,
+     .maxTextureDimension3D = WGPU_LIMIT_U32_UNDEFINED,
+     .maxTextureArrayLayers = WGPU_LIMIT_U32_UNDEFINED,
+     .maxDynamicUniformBuffersPerPipelineLayout = WGPU_LIMIT_U32_UNDEFINED,
+     .maxDynamicStorageBuffersPerPipelineLayout = WGPU_LIMIT_U32_UNDEFINED,
+     .maxSampledTexturesPerShaderStage = WGPU_LIMIT_U32_UNDEFINED,
+     .maxSamplersPerShaderStage = WGPU_LIMIT_U32_UNDEFINED,
+     .maxStorageBuffersPerShaderStage = WGPU_LIMIT_U32_UNDEFINED,
+     .maxStorageTexturesPerShaderStage = WGPU_LIMIT_U32_UNDEFINED,
+     .maxUniformBuffersPerShaderStage = WGPU_LIMIT_U32_UNDEFINED,
+     .maxUniformBufferBindingSize = WGPU_LIMIT_U64_UNDEFINED,
+     .maxStorageBufferBindingSize = WGPU_LIMIT_U64_UNDEFINED,
+     .minUniformBufferOffsetAlignment = WGPU_LIMIT_U32_UNDEFINED,
+     .minStorageBufferOffsetAlignment = WGPU_LIMIT_U32_UNDEFINED,
+     .maxVertexBuffers = WGPU_LIMIT_U32_UNDEFINED,
+     .maxVertexAttributes = WGPU_LIMIT_U32_UNDEFINED,
+     .maxVertexBufferArrayStride = WGPU_LIMIT_U32_UNDEFINED,
+     .maxInterStageShaderComponents = WGPU_LIMIT_U32_UNDEFINED,
+     .maxComputeWorkgroupStorageSize = WGPU_LIMIT_U32_UNDEFINED,
+     .maxComputeInvocationsPerWorkgroup = WGPU_LIMIT_U32_UNDEFINED,
+     .maxComputeWorkgroupSizeX = WGPU_LIMIT_U32_UNDEFINED,
+     .maxComputeWorkgroupSizeY = WGPU_LIMIT_U32_UNDEFINED,
+     .maxComputeWorkgroupSizeZ = WGPU_LIMIT_U32_UNDEFINED,
+     .maxComputeWorkgroupsPerDimension = WGPU_LIMIT_U32_UNDEFINED,
+ }
```

Fixes https://github.com/gfx-rs/wgpu-native/issues/214

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,22 +102,25 @@ lib-native-release: Cargo.lock Cargo.toml Makefile $(WILDCARD_SOURCE)
 	cargo build --release $(EXTRA_BUILD_ARGS)
 
 example-compute: lib-native examples/compute/main.c
-	cd examples/compute && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Debug .. $(GENERATOR_PLATFORM) && cmake --build .
+	cd examples/compute && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 .. $(GENERATOR_PLATFORM) && cmake --build .
 
 run-example-compute: example-compute
 	cd examples/compute && "$(OUTPUT_DIR)/compute" 1 2 3 4
 
 example-triangle: lib-native examples/triangle/main.c
-	cd examples/triangle && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Debug .. $(GENERATOR_PLATFORM) && cmake --build .
+	cd examples/triangle && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 .. $(GENERATOR_PLATFORM) && cmake --build .
 
-run-example-triangle: example-triangle
+example-triangle-release: lib-native-release examples/triangle/main.c
+	cd examples/triangle && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=1 .. $(GENERATOR_PLATFORM) && cmake --build .
+
+run-example-triangle-release: example-triangle-release
 	cd examples/triangle && "$(OUTPUT_DIR)/triangle"
 
 build-helper:
 	cargo build -p helper
 
 example-capture: lib-native build-helper examples/capture/main.c
-	cd examples/capture && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Debug .. $(GENERATOR_PLATFORM) && cmake --build .
+	cd examples/capture && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 .. $(GENERATOR_PLATFORM) && cmake --build .
 
 run-example-capture: example-capture
 	cd examples/capture && "$(OUTPUT_DIR)/capture"
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,9 @@ run-example-compute: example-compute
 example-triangle: lib-native examples/triangle/main.c
 	cd examples/triangle && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 .. $(GENERATOR_PLATFORM) && cmake --build .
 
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -102,22 +102,25 @@ lib-native-release: Cargo.lock Cargo.toml Makefile $(WILDCARD_SOURCE)
...
+
+run-example-triangle-release: example-triangle-release
 	cd examples/triangle && "$(OUTPUT_DIR)/triangle"
 
 build-helper:
 	cargo build -p helper
 
 example-capture: lib-native build-helper examples/capture/main.c
-	cd examples/capture && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Debug .. $(GENERATOR_PLATFORM) && cmake --build .
+	cd examples/capture && $(CREATE_BUILD_DIR) && cd build && cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 .. $(GENERATOR_PLATFORM) && cmake --build .
```
cwfitzgerald, line 123: Why do we need the exported commands?

src/conv.rs
```diff
@@ -436,8 +442,14 @@ pub fn map_image_copy_buffer(
...
         offset: native.offset,
-        bytes_per_row: NonZeroU32::new(native.bytesPerRow),
-        rows_per_image: NonZeroU32::new(native.rowsPerImage),
+        bytes_per_row: match native.bytesPerRow {
+            native::WGPU_COPY_STRIDE_UNDEFINED => None,
+            _ => NonZeroU32::new(native.bytesPerRow),
+        },
+        rows_per_image: match native.rowsPerImage {
+            native::WGPU_COPY_STRIDE_UNDEFINED => None,
+            _ => NonZeroU32::new(native.rowsPerImage),
```
cwfitzgerald, line 451: This feels like the wrong behavior for size = 0?

cwfitzgerald (changes requested):

Makefile
rajveermalviya: to make clangd + vscode work, not needed but doesn't hurt

src/conv.rs
rajveermalviya: (sorry for delay, I somehow missed this)

Wouldn't it be better to support both behaviors, it preserves backwards compatibility.
It requires `NonZero` types anyway or would you prefer unwrap-ing here and only support
new behavior.

cwfitzgerald: I would solely go based on what webgpu-headers expects - my hunch is that 0 for whole length isn't what it wants.

rajveermalviya: updated to only support new behavior

cwfitzgerald (approved):

almarklein:
I understand that this changes was done because other implementations seem to do things this way, but to be honest this approach feels rather fragile to me.

As an example, In #235 I update the `webgpu.h`, which introduces a new field in the `Limits`: the `maxBufferSize`. It took me about 2 hour today to figuring out that I also should [update the limit defaults](https://github.com/gfx-rs/wgpu-native/pull/235/commits/2f08880b8d4ee4ced00ad2dcde0045793537100e). The nasty thing about this is that (from how I understand this) every downstream application will have to update their default limits. I foresee many devs spending hours in frustration ...

And then there's the fact that there is `WGPU_LIMIT_U32_UNDEFINED` and `WGPU_LIMIT_U64_UNDEFINED`. If the type of a field is changed upstream (from u32 to u64, or back) this will cause a lot of pain downstream.

I'm not sure what the most elegant solution would be, but I imagine that it would help a lot if we had a function to create a struct populated with these undefined-placeholders. Then here in wgpu-native we could make sure to keep that up-to-date, and save pain downstream.

rajveermalviya:
There's also some discussion [here](https://github.com/webgpu-native/webgpu-headers/issues/158) about having the default struct macros to be included in `webgpu.h`. It would probably be best if that happens.

almarklein:
Ti work around this issue in wgpu-py I create the adapter, request its limits, and then use these as the default limits for creating devices.

PR #219, opened 2022-10-11
return `request{Adapter,Device}` error via callback

rajveermalviya:
don't panic on `request{Adapter,Device}` failure


cwfitzgerald (approved):
Nice!

Issue #220, opened 2022-10-27
Usage of GLOBAL with raw instance


guusw:
Is there any reasoning behind using the wgpu-core Global as a global variable instead of tracking it in a similar way that wgpu does, passing an Arc<Context> around?

I'm trying to integrate OpenXR in a project using wgpu-native which in some cases (Vulkan) adds additional requirements on the creation of the Instance (VkInstance)

However with the way wgpu-native works currently there's now way to use an external created instance other that making GLOBAL mutable (RwLock) or rewriting some stuff to assign the instance to created objects.

I'm thinking of adding `Arc<Global>` to the opaque C types to pass around the Global object instead of using the `GLOBAL` variable to support this case, I can probably open a PR when I'm done.

Let me know your thoughts or concerns.

rajveermalviya:
I too have a PR in pipeline which does something like this.

`wgpuCreateInstance` actually creates the instance, then in later functions instead of returning `Id` types directly we return a `Box::into_raw(Box::new())` of a struct wrapping the corresponding `Id` of the wgpu object, `Global`, and maybe error_callback (set via `wgpuDeviceSetUncapturedErrorCallback`) that gets passed down from device.

This removes all globally declared resources and removes all the various hacks we have currently.
Yes it's a big rewrite because each wgpu object has a corresponding struct declaration (same as in `wgpu-rs`). But as a result we get more error reporting instead of panics/expects, also ability to drop the instance (which hopefully fixes glfw segfault in #144).

I can send this at the end of this week or maybe sooner.

guusw:
Yeah I believe it's better for the library in general too, there's some other globals like ADAPTER_NAMES and SURFACE_TO_DEVICE that can be ridden of by just wrapping the required info in a boxed struct.

Anyway, I'm currently blocked by it for further integration with at least Vulkan and OpenXR so decided to start on it myself.
Not sure how far your PR is but here is at least my currently implementation of tracking the context for each handle => https://github.com/gfx-rs/wgpu-native/pull/221

Issue #154, opened 2021-12-20
[Question] How to release memory after use?


maxxnino:
I'm using wgpu-native in zig, and I noticed only some resource have destroyed function (Buffer, Texture, Device...). And many thing don't have destroy such as CommandBuffer, RenderpassEncoder, CommandEncoder. But they have Drop function instead, Texture and Buffer both have Destroy and Drop function. What should I use?

kvark:
The drop is the "normal" way of communicating that you don't need something.
For resource-based objects (textures and buffers) we also have "Destroy" to force the resource de-allocation. If you use destroy, you aren't able to use any other objects that depend on this one. If you just "Drop", the object allocation will internally be kept alive until all the dependents are gone.

maxxnino:
Thank for the clarify. After messing around, I found three drop function can be used:
 - `TextureView` for current `SwapChain` can be `Drop` after `SwapChainPresent` is called. Without this the memory steady increase about 100kb every 10 seconds in triangle example.
 - The same for `ShaderModule` after `DeviceCreateRenderPipeline` is called.
 - And `PipelineLayout` after `DeviceCreateRenderPipeline` is called.
 - 
But `CommandBuffer` and `CommandEncoder` can't use Drop function anywhere. This is why I'm really confuse

kvark:
Generally, you shouldn't need to drop command encoders/buffers. You create a command encoder, then you turn it into a command buffer, then you submit it. That's it.
Dropping them would be needed if you decide to no submit the work. So it's still needed (we might be missing these methods?) but generally useful only in rare scenarios.

almarklein:
Shouldn't it be possible to submit recorded commands multiple times? E.g. if you know that nothing has changed (for a particular part of the rendering).

cwfitzgerald:
wgpu doesn't support re-using command buffers like that, in general the situations where reusing command buffers is useful is minimal and ONE_TIME_SUBMIT or whatever it's called is a useful flag to have on cmdbuffers. That being said, renderbundles are how you are supposed to do save streams of draws to be replayed later.

almarklein:
Thanks for the clarification!

20kdc:
This is a useful issue to keep around, but I feel like this should be specified more directly, in documentation or something.
It's worth noting that https://gpuweb.github.io/gpuweb/ doesn't appear to say anything about single-submission, and even if it did, it's intended for JavaScript, so says nothing about dropping.
[...]

PR #221, opened 2022-10-27
Tie all created objects to their instance

guusw:
See https://github.com/gfx-rs/wgpu-native/issues/220


src/command.rs
```diff
@@ -22,14 +59,14 @@ pub unsafe extern "C" fn wgpuCommandEncoderFinish(
...
 
     let (id, error) =
-        gfx_select!(command_encoder => GLOBAL.command_encoder_finish(command_encoder, &desc));
+        gfx_select!(command_encoder => context.command_encoder_finish(command_encoder, &desc));
     if let Some(error) = error {
         // TODO figure out what device the encoder belongs to and call
         // handle_device_error()
         log::error!("command_encoder_finish() failed: {:?}", error);
-        None
+        null_mut()
```
rajveermalviya, line 67: nit: prefer `std::ptr::null_mut()` and remove import

src/lib.rs
```diff
@@ -197,32 +297,56 @@ macro_rules! map_enum {
 pub extern "C" fn wgpuCreateInstance(
     _descriptor: *const native::WGPUInstanceDescriptor,
 ) -> native::WGPUInstance {
-    // Rationale: See https://github.com/gfx-rs/wgpu-native/issues/116
-    // Because WGPUInstance is an opaque type this library controls and does not define the contents of, this is safe.
-    8 as native::WGPUInstance
+    let context = Context::new(
+        "wgpu",
+        wgc::hub::IdentityManagerFactory,
+        wgt::Backends::PRIMARY,
```
rajveermalviya, line 303: we should take backend via extras, same as `wgpuRequestAdapter`

src/device.rs
```diff
@@ -125,9 +135,9 @@ pub unsafe extern "C" fn wgpuAdapterGetProperties(
     adapter: native::WGPUAdapter,
     properties: &mut native::WGPUAdapterProperties,
 ) {
-    let adapter = adapter.expect("invalid adapter");
+    let (adapter, context) = unwrap_context_handle(adapter);
 
-    let maybe_props = gfx_select!(adapter => GLOBAL.adapter_get_info(adapter));
+    let maybe_props = gfx_select!(adapter => context.adapter_get_info(adapter));
     if let Ok(props) = maybe_props {
         properties.name = ADAPTER_NAMES
```
rajveermalviya, line 142: we can remove `ADAPTER_NAMES` global and instead return CString ptr and mem::forget it, making it user's responsibility to free the name string.

rajveermalviya (commented):
Thanks for doing this!
Since your PR is already here, we can go with it.
we can make error handling better in later PRs.

(haven't reviewed it thoroughly since it's just a draft, just added some points that I had planned in my PR)

src/device.rs
guusw: I moved the cached string into the adapter, could use mem::forget too but it hasn't been used like this (as a member of a returned structure) so I wanted to avoid that.

src/lib.rs
guusw: Added backend flags so you can indicate which backends are enabled by the hub

guusw:
I think it's ready for review

src/device.rs
rajveermalviya: make sense, seems like [dawn does](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/Adapter.cpp;l=89;drc=af07b3e7e43bd39dc707d11c3e32e12606c1d3bb) something similar.
Though we will need to add more strings later (description, architecture)

src/command.rs
```diff
@@ -1,18 +1,52 @@
...
+
 #[no_mangle]
 pub unsafe extern "C" fn wgpuCommandEncoderFinish(
     command_encoder: native::WGPUCommandEncoder,
     descriptor: Option<&native::WGPUCommandBufferDescriptor>,
 ) -> native::WGPUCommandBuffer {
-    let command_encoder = command_encoder.expect("invalid command encoder");
+    let (command_encoder, context) = {
+        command_encoder.as_ref().expect("invalid command encoder");
+        let b = Box::from_raw(command_encoder);
```
rajveermalviya, line 47: this will drop the `command_encoder` here, was it intended?

rajveermalviya (changes requested):

examples/triangle/main.c
```diff
@@ -71,6 +73,8 @@ int main() {
     return 1;
   }
 
+  instance = wgpuCreateInstance(&(WGPUInstanceDescriptor) {.nextInChain = NULL});
+
   WGPUSurface surface;
 
 #if WGPU_TARGET == WGPU_TARGET_MACOS
```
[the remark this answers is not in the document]
guusw: Ah missed that one.
Also indeed it should panic, will check that.

src/command.rs
guusw: I think I was slightly confused by this: https://github.com/gfx-rs/wgpu-native/issues/154 
In my code I was calling wgpuCommandEncoderFinish and pass the result to wgpuQueueSubmit and neither the encoder or command buffer had to be dropped. (now they would need to be since they return a box instead of just and id)

I'll change it so that you should explicitly call wgpuCommandEncoderDrop/wgpuCommandBufferDrop

rajveermalviya: > should explicitly call wgpuCommandEncoderDrop

in that case we should make `wgpuRenderBundleEncoderFinish` behave in same way, it will make drop-ing refs consistent.

guusw: I'm going of the example code and the linked, which seems to expect that command buffers/encoders are either used or dropped (not both)
So I unified the usage so that it works as follows, the following functions drop the corresponding objects:
```
WGPUComputePassEncoder:
- wgpuComputePassEncoderEnd
- wgpuComputePassEncoderDrop

WGPURenderPassEncoder:
- wgpuRenderPassEncoderEnd
- wgpuRenderPassEncoderDrop

WGPURenderBundleEncoder:
- wgpuRenderBundleEncoderFinish
- wgpuRenderBundleEncoderDrop

WGPURenderBundle:
- wgpuRenderBundleDrop

WGPUCommandEncoder:
- wgpuCommandEncoderFinish
- wgpuCommandEncoderDrop

WGPUCommandBuffer:
- wgpuQueueSubmit
- wgpuCommandBufferDrop
```

examples/triangle/main.c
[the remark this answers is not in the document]
guusw: Fixed this

rajveermalviya (approved):

cwfitzgerald (approved):
Cool!

PR #224, opened 2022-11-04
Feature/enumerate adapters

toloudis:
I would love some advice on how to proceed with this.

I am interested in potentially using wgpu-native as a cross-platform graphics api.  For my application, a volume rendering tool for scientific data (https://www.allencell.org/pathtrace-rendering.html), it makes sense to be able to support multi-gpu systems and potentially allow gpu selection by very custom means.  

So I have done a very basic first pass at exposing enumerate_adapters but the enumeration I get spans all possible backends.  Apart from checking against a previously created "compatible surface", is there some way to detect a current backend at run time?  Or is it reasonable to return adapters from all supported back ends?

Also api advice welcomed: for example, should there be a callback passed in to wgpuInstanceEnumerateAdapters?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -192,18 +192,23 @@ jobs:
           make example-capture
           make example-compute
           make example-triangle
+          make example-enumerate_adapters
       - if: ${{ matrix.run_examples }}
         name: Run examples debug
         run: |
           make run-example-capture
           make run-example-compute
+          make run-example-enumerate_adapters
       - name: Build examples release
         run: |
           make example-capture-release
           make example-compute-release
           make example-triangle-release
+          make example-enumerate_adapters-release
       - if: ${{ matrix.run_examples }}
         name: Run examples release
         run: |
           make run-example-capture-release
           make run-example-compute-release
+          make run-example-enumerate_adapters-release
+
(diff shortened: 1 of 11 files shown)


cwfitzgerald:
Sorry for the very long delay

> is there some way to detect a current backend at run time? 

Depends on what you mean by current. All supported backends are active on the instance at all times. Enumerate adapters can be further limited by the backend list.

I'll leave the C interface advice to @rajveermalviya

kushalkolar:
Wondering if this PR can receive any traction in the near future? It would be useful to select a specific GPU as the adapter in multi-GPUs systems. These can either be shared systems where multiple users login to a single multi-GPU system, and should ideally be able to choose a GPU based on what's available. Or on multi-GPU systems where some GPUs are being used for compute (CUDA, training neural networks, etc.), and other GPUs are used for visualization.

Similar to the author of this PR, we are also using it for scientific visualization via `wgpu-py`: https://github.com/kushalkolar/fastplotlib/

rajveermalviya:
Apologies for the long delay, somehow I missed this PR.

In order to get this merged ASAP, I have rebased and made some changes fixing the possible memory leak.
Also added an mini example demonstrating the use.

As for the C API, double call seems fine as this will probably be called only once in an application.

kushalkolar:
Wow, you guys are epic :smile: !

almarklein (approved):

toloudis:
Thank you!

PR #226, opened 2022-11-09
Add Silk.NET to binding list

Beyley:
Silk.NET now has raw WebGPU bindings for C# (as opposed to wgpu.NET, which acts a lot more as a wrapper)


cwfitzgerald (approved):
Thanks!

PR #229, opened 2022-11-21
Consistent libname

### LINKED ISSUE #228: Inconsistent DLL naming in Windows release
Since [v0.10.4.1](https://github.com/gfx-rs/wgpu-native/releases/tag/v0.10.4.1), the dll provided the Windows build is called `libwgpu.dll`, whereas it was called `wgpu_native.dll` before.

However, the compagnon .lib file (which was missing in v0.10.4.1 but has been added back), still refers to `wgpu_native.dll` internally, as can be seen in the "DLL name" entry of all symbols listed by `dumpbin -headers libwgpu.lib`:

```
>dumpbin -headers libwgpu.lib
Microsoft (R) COFF/PE Dumper Version 14.29.30142.1
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file libwgpu.lib
[...]
  Version      : 0
  Machine      : 8664 (x64)
  TimeDateStamp: F24E255C
  SizeOfData   : 00000030
  DLL name     : wgpu_native.dll
  Symbol name  : wgpu_render_bundle_set_pipeline
  Type         : code
  Name type    : name
  Hint         : 99
  Name         : wgpu_render_bundle_set_pipeline
[...]
```

Possible workarounds:

 - Rename the dll back to wgpu_native.dll
 - Regenerate the .lib file with this kind of trick: https://stackoverflow.com/questions/9360280/how-to-make-a-lib-file-when-have-a-dll-file-and-a-header-file

**NB** Library names starting with "lib" is unix-style but not so windows-style, I think wgpu_native.dll was ok, wgpu.dll would be even better imho.

edit: I see that the symbol `wgpuAdapterEnumerateFeatures` is missing from the DLL

almarklein:
Probably closes #228

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -40,7 +40,7 @@ jobs:
       IMAGE: manylinux_2_24_x86_64
     steps:
     # Common part (same for nearly each build)
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         submodules: true
     - name: Set WGPU_NATIVE_VERSION
@@ -67,7 +67,7 @@ jobs:
           docker rm $CID
     # Upload (same for each build)
     - name: Upload
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         path: dist
         name: dist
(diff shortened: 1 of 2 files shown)


almarklein:
Also tried to get rid of some warnings by updating action versions, but this does not seem to help. As far as I can see, the warnings originate from the actions we use, not our own workflow, so I guess its a matter of waiting for upstream actions to update.

PR #233, opened 2023-01-11
Dont rename libs for packaging

### LINKED ISSUE #231: Packaging fails due to incorrect lib name
Had this issue locally and verified that it also happens in the latest github actions run: https://github.com/gfx-rs/wgpu-native/actions/runs/3557239659/jobs/5975115486#step:4:995

The problem seems to be that the Makefile is using a `LIB_NAME` variable that doesn't match the output from cargo: https://github.com/gfx-rs/wgpu-native/blob/master/Makefile#L46

almarklein:
Closes #231

The binary libs that we package as part of a release were renamed to make the names consistent across OS's. Apparently this had unintended side-effects, see #228. I tried to fix this in #229 but that just broke the packaging instead 😕 . This fixes things again, simply by not renaming the libs.

I tried making the build produce `libwgpu_native.dll` instead of `wgpu_native.dll`, but was not able to make that work. If someone knows an easy way to add the "lib" prefix on Windows, that'd be great, but I have a feeling this might be muddy ... 

Anyway, downstream packages might need to update some paths:
* libwgpu.dll -> wgpu_native.dll
* libwgpu.so -> libwgpu_native.so
* libwgpu.dylib -> libwgpu_native.dylib
*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,6 @@ FFI_DIR:=ffi
 BUILD_DIR:=build
 CREATE_BUILD_DIR:=
 OUTPUT_DIR:=
-FINAL_LIB_NAME:=libwgpu
-
 
 WILDCARD_SOURCE:=$(wildcard src/*.rs)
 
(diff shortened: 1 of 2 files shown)


rajveermalviya (commented):
can we also include static libs?
iirc it was requested in matrix chat.

rajveermalviya (approved):

almarklein:
> can we also include static libs?

Done

radgeRayden:
I was going to mention in the other issue that renaming has consequences when linking a different shared library to webgpu, but didn't have the details on hand to back it up. Either way this solution solves these other concerns as well.

almarklein:
Yeah, in hindsight renaming was a bad idea.

PR #239, opened 2023-01-29
support nullable args in `requestAdapter` & `requestDevice`

rajveermalviya:
As per the headers:

- [`options` arg is nullable for `wgpuInstanceRequestAdapter`](https://github.com/webgpu-native/webgpu-headers/blob/fa1c6ab4927ef1fa5731907b42b62ea93119866c/webgpu.h#L1497)
- [`descriptor` arg is nullable for `wgpuAdapterRequestDevice`](https://github.com/webgpu-native/webgpu-headers/blob/fa1c6ab4927ef1fa5731907b42b62ea93119866c/webgpu.h#L1413)

### DIFF
--- a/examples/capture/main.c
+++ b/examples/capture/main.c
@@ -28,33 +28,16 @@ int main(int argc, char *argv[]) {
   int width = 100;
   int height = 200;
 
-  WGPUInstance instance = wgpuCreateInstance(&(WGPUInstanceDescriptor) {.nextInChain = NULL});
+  WGPUInstance instance =
+      wgpuCreateInstance(&(WGPUInstanceDescriptor){.nextInChain = NULL});
 
   WGPUAdapter adapter;
-  wgpuInstanceRequestAdapter(instance,
-                             &(WGPURequestAdapterOptions){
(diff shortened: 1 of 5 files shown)


almarklein (approved):
Ah, that makes things easier already. 

The trick I ended up using in wgpu-py is to create the adapter, request its limit, and use these as the default limits for the device. (Will also post this on the corresponding issue, for future reference.) -> https://github.com/gfx-rs/wgpu-native/pull/218

rajveermalviya:
@almarklein can this be included in a new release?

(also what's the process of release in general, would creating a new tag be enough?)

almarklein:
Sure!

> (also what's the process of release in general, would creating a new tag be enough?)

Yes, you simply pass a tag called `vA.B.C.D` and CD does the rest. The idea is to follow the version of wgpu-core and then add a sequence number. So `0.14.2.4` is up next, I think.

rajveermalviya:
Cool!, created [v0.14.2.4](https://github.com/gfx-rs/wgpu-native/releases/tag/v0.14.2.4)

PR #241, opened 2023-02-01
fix segfaults caused by null values

rajveermalviya:
when users passes `NULL` for pointer args in some functions, because of no null checks in place - library segfaults and crashes without any information printed.
now we panic with an error message instead, which should be better.


almarklein (approved):
❤️

PR #245, opened 2023-02-26
drop cmdbuffer handles in `wgpuQueueSubmitForIndex`

rajveermalviya:
currently command buffer handle (boxed struct wrapping actual `CommandBuffer` & `Arc<Context>`) isn't dropped in `wgpuQueueSubmitForIndex`, so basically leaking a ref of `Arc<Context>`. And since `context.queue_submit()` drops the actual `CommandBuffer` internally, calling `wgpuCommandBufferDrop` after `wgpuQueueSubmitForIndex` is useless as it will panic complaining the actual `CommandBuffer` it's already vacant.

so this pr fixes it, completely consuming command buffer on queue submit.

this behavior is already correct in `wgpuQueueSubmit`, just the wgpu extension function `wgpuQueueSubmitForIndex` had this bug.

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -1011,7 +1011,7 @@ pub fn map_swapchain_descriptor(
                 native::WGPUCompositeAlphaMode_Inherit => wgt::CompositeAlphaMode::Inherit,
                 _ => panic!("invalid alpha mode for swapchain descriptor"),
             },
-            unsafe { make_slice(extras.viewFormats, extras.viewFormatCount as usize) }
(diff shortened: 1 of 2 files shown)


almarklein (approved):

almarklein (approved):

almarklein:
Out of curiosity, what is the effect of this change? Does this solve a particular issue?

rajveermalviya:
apologies i have updated the description to explain more.

PR #246, opened 2023-03-05
add stubs for unimplemented fns & implement some fns

rajveermalviya:
export all the fns present in the headers. this makes some binding generators that check `.so`/`.a` happy; no more complains for missing fn signatures.

also it's probably makes it better for us to track which functions are unimplemented.

new fns implemented are:
- `wgpuCommandEncoderResolveQuerySet`
- `wgpuCommandEncoderWriteTimestamp`
- `wgpuComputePassEncoderBeginPipelineStatisticsQuery`
- `wgpuComputePassEncoderEndPipelineStatisticsQuery`
- `wgpuDeviceCreateQuerySet`
- `wgpuRenderPassEncoderBeginPipelineStatisticsQuery`
- `wgpuRenderPassEncoderEndPipelineStatisticsQuery`
- `wgpuBufferGetConstMappedRange`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -35,7 +35,6 @@ fn main() {
         .header("ffi/webgpu-headers/webgpu.h")
         .header("ffi/wgpu.h")
         .parse_callbacks(Box::new(bindgen::CargoCallbacks))
-        .blocklist_type("(^WGPUProc).*")
         .blocklist_function("wgpuGetProcAddress")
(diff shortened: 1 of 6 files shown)


almarklein (approved):
Nice! One step closer to "completion". Good idea to put the unimplemented functions in a separate module 👍

PR #248, opened 2023-03-13
allow 0 for `BufferBindingLayout.minBindingSize`

rajveermalviya:
as per the spec, `GPUBufferBindingLayout.minBindingSize` should allow 0 as default value instead of `WGPU_WHOLE_SIZE`. 
See https://gpuweb.github.io/gpuweb/#dom-gpubufferbindinglayout-minbindingsize

(also thanks to @eliemichel for catching this)

(this is breaking change, as current code using `WGPU_WHOLE_SIZE` will result in UB)


kvark:
Can we put an `assert_ne!(size, WHOLE_SIZE)` to prevent the unexpected?

rajveermalviya:
added an assert

almarklein (approved):

PR #249, opened 2023-03-23
Add wgpu_native.dll.lib and wgpu_native.pdb to windows package

### LINKED ISSUE #180: Win64 release and debug binaries are missign the *.dll.lib stubs for linking
Hello, 

The latest Windows WGPU native pre-build binaries are missing the *.lib stubs for linking against. 

This was was an that existed in the past but was fixed a few versions back, but has now returned. 

Without them being part of the release package, the only other way to use the WGPU native DLLs is to download and build the source locally. 

Thank you for all the amazing work,
- Chris  (d.m.p)

BenjaminSchaaf:
The `wgpu_native.dll.lib` file is the import library for the `dll`, without it you can't actually link against the dynamic library.

The `wgpu_native.pdb` file contains debug information.

Fixes #180

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ package: lib-native lib-native-release
 		rm -f dist/$$ARCHIVE; \
 		sed 's/webgpu-headers\///' ffi/wgpu.h > wgpu.h ;\
 		if [ $(OS_NAME) = windows ]; then \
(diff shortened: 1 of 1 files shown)


rajveermalviya (approved):

almarklein:
What's the difference between the `.lib`  and the `.dll.lib`?

PR #251, opened 2023-04-11
Generate CMake distribution in CD

eliemichel:
Hello,

I am trying to make the integration of the precompiled binaries in a **CMake**-based project as easy and **portable** as possible for my [WebGPU C++ guide](https://eliemichel.github.io/LearnWebGPU/getting-started/hello-webgpu.html) and I came up with a solution that satisfies me:

 - Easy to integrate, with a `webgpu` CMake target to link like any library and that manages the pre-compiled binaries
 - Define a `WEBGPU_BACKEND_WGPU` to distinguish this distribution for the Dawn-based alternative that I also provide
 - Follow modern idioms of CMake
 - Works with emscripten

Currently I point my readers to my [WebGPU-distribution](https://github.com/eliemichel/WebGPU-distribution) repository but I think this is something that would be worth making available directly in the release of `wgpu-native`.

This PR adds to the CD process the generation of a *distribution* of WebGPU that can be simply copied in a CMake project and integrated with `add_subdirectory(webgpu)`. More info to be found in [distribution/REAMD.md](https://github.com/eliemichel/wgpu-native/blob/feature-cmake-distribution/distribution/README.md).

This is something that I am going to also discuss with Dawn so that ideally we end up with a common way of packaging binaries for CMake. I'll be happy to discuss the details and adapt to as many use cases as possible!


.github/workflows/cd.yml
```diff
@@ -271,16 +271,86 @@ jobs:
...
+          unzip dist/wgpu-*-debug.zip distribution/bin/
+          unzip -d distribution/bin $zipfile *.so *.dylib *.dll
+        done
+        mv distribution webgpu
+        zip -r wgpu-distribution.zip webgpu
+    - name: Upload
+      uses: actions/upload-artifact@v3
+      with:
+        path: wgpu-distribution.zip
+        name: dist-cmake
```
almarklein, line 340: Won't the zip-files for the debug and release distribution overwrite each-other?

Also, I think you can simply use `name: dist`, which keeps the GH release job simpler.

distribution/README.md
```diff
@@ -0,0 +1,69 @@
+WebGPU distribution
+===================
+
+This is pre-compiled distribution of [WebGPU native](https://github.com/webgpu-native/webgpu-headers) based on [`wgpu-native`](https://github.com/gfx-rs/wgpu-native) and ready to be used in a [CMake](https://cmake.org/) project.
```
almarklein, line 4:
```suggestion
This is a pre-compiled distribution of [WebGPU native](https://github.com/webgpu-native/webgpu-headers) based on [`wgpu-native`](https://github.com/gfx-rs/wgpu-native) and ready to be used in a [CMake](https://cmake.org/) project.
```

almarklein (commented):
Thanks! This sounds like it would be useful, but (being mostly a Python dev) I cannot judge whether this approach is good practice or something like that.

.github/workflows/cd.yml
eliemichel: Indeed, fixed!

.github/workflows/cd.yml
```diff
@@ -271,16 +271,86 @@ jobs:
...
+          mkdir -p distribution/bin/$arch
+          unzip dist/wgpu-*-debug.zip distribution/bin/
+          unzip -d distribution/bin $zipfile *.so *.dylib *.dll
+        done
+        mv distribution webgpu
+        zip -r wgpu-distribution-debug.zip webgpu
+    - name: Upload
+      uses: actions/upload-artifact@v3
+      with:
+        path: wgpu-distribution.zip
```
almarklein, line 339:
```suggestion
        path: wgpu-distribution-debug.zip
```

almarklein (commented):
missed a spot :)

eliemichel:
Oh thanks my bad!

almarklein:
This is open for a while even though this is about finished. I think this needs a round of review from someone with more experience (than me) in build systems. cc @rajveermalviya

waywardmonkeys:
Is there something I can do to help push this along?

waywardmonkeys:
@rajveermalviya Any chance I can help get this wrapped up for the next release?

rajveermalviya:
(Apologies for missing this)

Tbh instead of this, I would prefer a setup where user can clone the repo (or github's download zip) to a directory and then import that directory via `add_subdirectory`.

Then the `CMakeLists.txt` at the root in this repo could the have an option like `DOWNLOAD_PREBUILT_BINARIES`, where default would be false. So, the binaries would be built from source by default & if that option is enabled then the binaries would be downloaded from Github Releases. Such a setup would also allow us to mirror cargo optional features that are currently disabled in the release packages.

But not really sure how emscripten would fit in that.

eliemichel:
I agree actually, not my topmost priority at the moment but I'll work on this eventually!

waywardmonkeys:
`wasmtime` does something similar as well. I have it on my list to try to get to this sooner rather than later.

waywardmonkeys:
I have a working prototype of this locally now. I'll need #359 to land and then I can start on a series of PRs that get us to (what I hope is) a good place.

cwfitzgerald:
We definitely still want this, but we dropped the ball on reviews. I'm going to close this, we definitely want integration with cmake somehow.

PR #252, opened 2023-04-13
refactor examples

rajveermalviya:
- examples:
    - all the examples are refactored to use `Drop()` functions and correctly do cleanup
    - shader loading previously leaked memory [here](https://github.com/gfx-rs/wgpu-native/blob/240e65a96e20f4ee409c70ed45c030534b6f7100/examples/framework.c#L15) & [here](https://github.com/gfx-rs/wgpu-native/blob/240e65a96e20f4ee409c70ed45c030534b6f7100/examples/framework.c#L22)
    - remove helper rust library for `capture` example, instead [`stb_image_write.h`](https://github.com/nothings/stb/blob/master/stb_image_write.h) header is vendored and used for writing png image.
    - cmake build system is completely refactored, it now has a toplevel `examples/CmakeLists.txt` and projects included via `add_subdirectory()`.
    - glfw dependency is now included directly as a git submodule under `examples/vendor/glfw`, and compiled as a static library, this removes the need for setting up dependency loading on both ci & when running locally. And also allows us to test the experimental wayland support on linux.

- `ci.yml`:
    - now separated into 3 jobs only `integrity-build`, `clippy`, `examples`
    - clippy job runs clippy for all platforms + atleast one of both 32bit & 64bit arch combinations
    - examples job builds the examples on possible platforms & runs them (on macos we only build, can't run without metal in vm).
    - no longer tests with rust nightly, if required then I can add it back in a separate pr.


.github/workflows/ci.yml
```diff
@@ -8,178 +8,202 @@ on:
...
+        uses: dtolnay/rust-toolchain@stable
         with:
-          submodules: 'true'
-      - run: rustup component add clippy
-      - run: cargo clippy
+          target: ${{ matrix.target }}
+      - name: Setup Environment
+        run: ${{ matrix.setup_env }}
+      - name: Run tests
+        run: cargo test
```
almarklein, line 189: So this does not only run the examples. What does `cargo test` do, anyway?

almarklein (approved):
This looks good to me. I haven't checked the refactoring of makefiles and C code in detail, but if CI agrees, then I guess it's good 👍

.github/workflows/ci.yml
rajveermalviya: it runs some tests that are generated by `bindgen`

https://github.com/gfx-rs/wgpu-native/actions/runs/4687763977/jobs/8307456351#step:7:168

rajveermalviya:
seems like github is waiting for the ci tasks that no longer exists, also I can't force merge the PR.
cc @cwfitzgerald

kvark:
@almarklein we need a force merge here. Are you happy with the code? I haven't reviewed it, just want to make sure you can force merge if you want to. Will help updating the branch protection rules if needed.

almarklein:
> Will help updating the branch protection rules if needed.

I can't merge either. Only "Enable auto-merge". So if you could update the branch protection rules, that'd be great 👍

rajveermalviya:
@kvark @cwfitzgerald ping

PR #257, opened 2023-04-27
Fill in more of `WGPUAdapterProperties`

PyryM:
`.driverDescription`: filled in with a combination of the driver name and info, e.g., "NVIDIA  531.41" 
`.vendorName`: empty string (rather than null pointer)
`.architecture`: empty string (rather than null pointer)

Looking at Dawn and Emscripten's implementations, there does not seem to be much consistency in how missing fields are handled: Dawn seems to return empty strings, while Emscripten returns null pointers. My feeling is that empty strings are more in line with the JS WebGPU API which explicitly returns empty strings (and also safer if someone printf's these fields without checking whether they are null).

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -67,6 +67,9 @@ pub unsafe extern "C" fn wgpuInstanceRequestAdapter(
                     context: context.clone(),
                     id: adapter,
                     name: CString::default(),
+                    vendor_name: CString::default(),
+                    architecture_name: CString::default(),
+                    driver_desc: CString::default(),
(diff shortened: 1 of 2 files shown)


rajveermalviya (approved):
Seems fine.

it's all confusing because `webgpu.h` doesn't have 1-1 fields matching webgpu js spec. ^^
`webgpu.h` needs a complete makeover as the webgpu spec is almost finalized.

Thanks!

almarklein (approved):

PR #258, opened 2023-04-30
Add `Force32` guard to `WGPUNativeFeature`

eliemichel:
Already present in other structs, it must have been forgotten in `WGPUNativeFeature`.

By the way it would match more closely the standard `webgpu.h` header to call this struct `WGPUNativeFeatureName`, and have the values look more like `WGPUNativeFeaturePushConstants` than `WGPUNativeFeature_PUSH_CONSTANTS`.

Side note: Using the `Native` prefix is not as informative as if it would use a name more specific to this very WebGPU backend (as opposed to "Dawn") but I guess this is a general problem of the name "`wgpu-native`".

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -22,7 +22,7 @@ typedef enum WGPUNativeFeature {
     WGPUNativeFeature_MULTI_DRAW_INDIRECT = 0x60000003,
     WGPUNativeFeature_MULTI_DRAW_INDIRECT_COUNT = 0x60000004,
     WGPUNativeFeature_VERTEX_WRITABLE_STORAGE = 0x60000005,
-    WGPUNativeSType_Force32 = 0x7FFFFFFF
+    WGPUNativeFeature_Force32 = 0x7FFFFFFF
 } WGPUNativeFeature;
 
 typedef enum WGPULogLevel {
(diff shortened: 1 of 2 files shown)


ffi/wgpu.h
```diff
@@ -22,6 +22,7 @@ typedef enum WGPUNativeFeature {
     WGPUNativeFeature_MULTI_DRAW_INDIRECT = 0x60000003,
     WGPUNativeFeature_MULTI_DRAW_INDIRECT_COUNT = 0x60000004,
     WGPUNativeFeature_VERTEX_WRITABLE_STORAGE = 0x60000005,
+    WGPUNativeSType_Force32 = 0x7FFFFFFF
```
rajveermalviya:
```suggestion
    WGPUNativeFeature_Force32 = 0x7FFFFFFF
```

eliemichel: My bad indeed, thx!

rajveermalviya:
I am fine with changing enums to completely follow `PascalCase`-ing. 
can you also make that change?

pinging @almarklein for their opinion.

almarklein:
I'll go with whatever's more consistent :)

almarklein (approved):

eliemichel:
@rajveermalviya I made the switch to PascalCase!

rajveermalviya (approved):
Thanks!

PR #262, opened 2023-05-15
refactor

rajveermalviya:
removed some unnecessary meta-programming, this makes some functions little more verbose but improves readability & makes it simpler to understand for new contributors.

some additional changes:
- replace some `unwrap` calls with `expect('msg')`
- replace usages of `slice::from_raw_parts` + `copy_from_slice` with direct call to [`std::ptr::copy_nonoverlapping`](https://doc.rust-lang.org/std/ptr/fn.copy_nonoverlapping.html)
- generate error msg for uncaptured error callback using `err.to_string()` instead of `format!("{err}")`, this makes string human readable.
- remove some redundant `drop` calls
- make some mapping functions `#[inline]`
- remove `lazy_static` dependency

### Testing
tested via running examples on linux

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -909,7 +909,6 @@ name = "wgpu-native"
 version = "0.0.0"
 dependencies = [
  "bindgen",
- "lazy_static",
  "log",
  "naga",
  "paste",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,6 @@ rev = "b99d58ea435090e561377949f428bce2c18451bb"
 version = "0.12.0"
 
 [dependencies]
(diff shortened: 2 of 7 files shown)


almarklein (approved):
My knowledge of Rust is insufficient to judge these changes, but if the compilers agree and the examples run (they also run for me, on MacOS), I am 👍

rajveermalviya:
> - generate error msg for uncaptured error callback using err.to_string() instead of format!("{err}"), this makes string human readable.

@almarklein note that with this change, you may have to update the shader error parsing in wgpu-py

almarklein:
Thanks for the heads-up! I made a note about it.

PR #264, opened 2023-05-21
Better error handling

rajveermalviya:
(even though it is divided in ~~8 commits~~ 9 commits apologies for a large PR, but I think it had to be done)

### Error handling
All the errors are now handled in some way, either storing them to be captured by `wgpuDevice{PopErrorScope,PushErrorScope}`, calling the uncaptured error handler or panic-ing if it's a fatal error.
- Fatal error operations are same ones in [wgpu-rs](https://github.com/search?q=repo%3Agfx-rs%2Fwgpu++path%3Awgpu%2Fsrc%2Fbackend%2Fdirect.rs+self.handle_error_fatal%28&type=code).
- Some errors are ignored for example in `wgpuBufferDestroy`, it's for following the spec (they are documented as such in code) also wgpu-rs deals with them in the same way.

most of the error handling code is *yoinked* from wgpu-rs, this results in consistent error handling mechanism, and consistent error messages (much better than before). Error messages now include the function name & the wgpu object label to better identify the problems.

### Refactoring
All the `extern "C"` functions now live in `lib.rs`, this does create a large single file but it makes finding functions little bit easier. wgpu-rs does the same see: [wgpu/src/backend/direct.rs](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu/src/backend/direct.rs)

### Misc
- sort drop functions defs in `wgpu.h`
- add `wgpuQueueDrop`
- implement `wgpuDevicePopErrorScope` & `wgpuDevicePushErrorScope` 

### Testing
tested by running all the examples on linux.


almarklein (approved):
Nice!

PR #269, opened 2023-06-27
Fix renamed wgpu symbols

guusw:
Seems like they moved some wgpu symbols (HubReport, GlobalReport, Global, IdentityManagerFactory)
Renamed them so wgpu-native can compile with latest.

I needed these fixes to get wgpu working on the trunk branch (works on Vision now)
I'm not sure how to deal with versions since this is targeting trunk on the wgpu repository and probably doesn't have a release & version number yet, does wgpu-native use the staging branch for this?

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -895,7 +895,7 @@ pub fn map_stencil_face_state(
 }
 
 #[inline]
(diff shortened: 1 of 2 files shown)


rajveermalviya:
Currently, we don't have a staging branch that targets `wgpu` trunk directly.
But for now #282 updates `wgpu`.

PR #271, opened 2023-06-28
Implement RG11B10UfloatRenderable

AlbinBernhardssonARM:
This was implemented in wgpu-core 16.0 so we just need to expose it.


rajveermalviya (approved):
Thanks!

almarklein (approved):

PR #279, opened 2023-07-06
update wgpu-core to 0.16.1

rajveermalviya:


rajveermalviya:
pushing through because important for the release, and it seems indisputable

PR #284, opened 2023-07-24
fix macos arm64 build

rajveermalviya:
### Context
Currently macos arm64 fails to set the correct `MACOSX_DEPLOYMENT_TARGET`:
https://github.com/gfx-rs/wgpu-native/actions/runs/5643538134/job/15285583834#step:6:1

- remove `SDKROOT`: I am guessing it was needed before because building for Apple Silicon required beta xcode, which is not necessary anymore (I looked at [rust-analyzer's release action](https://github.com/rust-lang/rust-analyzer/blob/master/.github/workflows/release.yaml)).
- for macos arm64 build set `MACOSX_DEPLOYMENT_TARGET` to `11.0` (first release to support Apple Silicon)
- don't set the xcode version manually, use whatever's default from the runner-image.

### Testing
[resulting builds not tested](https://github.com/rajveermalviya/wgpu-native/actions/runs/5645369894), action was tested in my fork.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -154,7 +154,7 @@ jobs:
     env:
       TARGET: x86_64-apple-darwin
       ARCHIVE_NAME: wgpu-macos-x86_64
-      MACOSX_DEPLOYMENT_TARGET: 10.13
+      MACOSX_DEPLOYMENT_TARGET: "10.13"
     steps:
     # Common part (same for each build)
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


almarklein (approved):
I cannot extensively test this, since wgpu-py is waay behind, but the lib does load and can report its version.

rajveermalviya:
forgot to tag the release, [tagged now](https://github.com/gfx-rs/wgpu-native/tags)

Issue #290, opened 2023-09-04
Non-compliant handling of LoadOp/StoreOp


eliemichel:
Currently, `wgpu-native` always requires `depthLoadOp`, `depthStoreOp`, `stencilLoadOp` and `stencilStoreOp` to be defined. Setting either of these to `Undefined` (which is a value that exists in `webgpu.h`) raises a runtime exception, e.g.:

```
thread '<unnamed>' panicked at 'Unknown load op: 0', src\conv.rs:9:1
```

However, according to the spec's [GPURenderPassDepthStencilAttachment Valid Usage](https://www.w3.org/TR/webgpu/#abstract-opdef-gpurenderpassdepthstencilattachment-gpurenderpassdepthstencilattachment-valid-usage), the value `Undefined` MUST be used in some cases:
 - When `format` has no depth aspect or `depthReadOnly` is true, `depthLoadOp` and `depthStoreOp` MUST be undefined.
 - When `format` has no stencil aspect or `depthReadOnly` is true, `stencilLoadOp` and `stencilStoreOp` MUST be undefined.

As of now, Dawn implements the specified behavior so it raises an error when say `stencilLoadOp` is not undefined but no stencil is used, and inversely in that case wgpu-native raises an error if `stencilLoadOp` is not undefined. I end up with this code in a lot of places:

```C++
#ifdef WEBGPU_BACKEND_WGPU
	depthStencilAttachment.stencilLoadOp = LoadOp::Clear;
	depthStencilAttachment.stencilStoreOp = StoreOp::Store;
#else
	depthStencilAttachment.stencilLoadOp = LoadOp::Undefined;
	depthStencilAttachment.stencilStoreOp = StoreOp::Undefined;
#endif
```

An **easy fix** would be to update `conv.rs` to just map `Undefined` to any non-undefined value. A "cleaner" way would be to use `std::option` in `wgpu-core` in `struct RenderPassDepthStencilAttachment`, and propagate these changes throughout the code base. Which one would you advise?

almarklein:
I'd say we could implement the easy fix first, so we're at least compatible with Dawn, and behave correctly when "correct" inputs are given. On the long run the cleaner solution would (obviously) be preferred but that involves changes in multiple repos.

Are you perhaps up for a PR to implement the easy fix?

eliemichel:
Yes, here it is! #292

PR #292, opened 2023-09-12
Allow Undefined value for load/store op

eliemichel:
This is the "easyfix" for #290 by wrapping the `map_load_op` and `map_store_op` conversion utils so that instead of panic-ing they return an arbitrary value.

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -27,11 +27,11 @@ map_enum!(
     Store
 );
 #[inline]
-pub fn map_load_op(value: i32) -> wgc::command::LoadOp {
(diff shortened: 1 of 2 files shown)


almarklein (approved):

almarklein:
LGTM. CI shows some type errors to resolve though.

eliemichel:
Indeed, I fixed the errors, as well as the comment that led me to the erronous signature ;)

almarklein (approved):

rajveermalviya (approved):
LGTM, thanks!

PR #294, opened 2023-09-16
New enum values for wgpu-native specific enums

rajveermalviya:
### Context
In webgpu-native meeting, it was decided to use top 16 bits as a namespace for enum values.
https://github.com/webgpu-native/webgpu-headers/issues/214

### Testing
Ran the examples on macos

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -5,21 +5,21 @@
 
 typedef enum WGPUNativeSType {
     // Start at 6 to prevent collisions with webgpu STypes
-    WGPUSType_DeviceExtras = 0x60000001,
-    WGPUSType_RequiredLimitsExtras = 0x60000002,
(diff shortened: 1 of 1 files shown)


rajveermalviya:
(keeping it as draft - because it's stacked on top of #293, so that one should be merged first)

ffi/wgpu.h
```diff
@@ -5,21 +5,21 @@
 
 typedef enum WGPUNativeSType {
     // Start at 6 to prevent collisions with webgpu STypes
```
almarklein:
```suggestion
    // Start at 0003 since that's allocated range for wgpu-native
```

almarklein (commented):
Let's fix the comment, but otherwise this LGTM.

ffi/wgpu.h
rajveermalviya: Ahh, nice find.

PR #296, opened 2023-09-18
Timestamp queries in render/compute pass descriptor

eliemichel:
This is an attempt at adding support for the `timestampWrites` field of render and compute pass descriptors.

Notable remarks:
 - The API for timestamp writes in `wgpu-core` **does not match the standard** `webgpu.h`. In particular, it does not support different query sets in the same descriptor, nor multiple queries for the same timestamp location (although I admit there must be only very limited use case for these).
 - Resolving queries that have not been written leads to **crash** with a lost device with no error message at all (already the case prior to this PR).
 - The writing/resolving multiple times the same query suspicioulsy returns the same value all the time in [my basic experiment](https://github.com/eliemichel/LearnWebGPU-Code/blob/step095-timestamp-queries/Application.cpp) ([doc](https://eliemichel.github.io/LearnWebGPU/advanced-techniques/benchmarking/time.html)) (running on DX12 backend).
 - I do **not check** that the query set has **the correct *timestamp* type**, because none of the other functions manipulating query sets does so, hence I assume `wgpu-core` performs the check.
 - Not sure how idiomatic/minimal my code is, I am open to suggestion and improvment of my rust style ;)

(**Bonus:** This PR adds support for the new `gles3_minor_version` instance option.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1012,11 +1012,9 @@ pub unsafe extern "C" fn wgpuCommandEncoderBeginComputePass(
     }).flatten();
 
     let desc = match descriptor {
-        Some(descriptor) => {
-            wgc::command::ComputePassDescriptor {
-                label: ptr_into_label(descriptor.label),
-                timestamp_writes: timestamp_writes.as_ref(),
-            }
+        Some(descriptor) => wgc::command::ComputePassDescriptor {
+            label: ptr_into_label(descriptor.label),
+            timestamp_writes: timestamp_writes.as_ref(),
         },
         None => wgc::command::ComputePassDescriptor::default(),
     };
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -965,9 +965,68 @@ pub unsafe extern "C" fn wgpuCommandEncoderBeginComputePass(
         )
     };
 
+    let all_timestamp_writes = descriptor.map(|descriptor| {
+        make_slice(descriptor.timestampWrites, descriptor.timestampWriteCount)
+        .iter()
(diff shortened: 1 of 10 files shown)


eliemichel:
Updated to match the new API for timestamp writes in `webgpu.h` (see [this change](https://github.com/webgpu-native/webgpu-headers/commit/b760e9e09da78e3d10b4cf5a8cdc9264ff74d08f)).

Not sure why CI fail on MacOS

rajveermalviya:
> Not sure why CI fail on MacOS

seems like cargo having some problems with wgpu-hal dep, I tested with this rev locally & I can reproduce it.
Changing it to current trunk of wgpu works fine, so let's just change it to that. (This is fine because, this is going to be released in 0.18, when wgpu does 0.18)

rajveermalviya:
Need a rebase or merge trunk

eliemichel:
Done ;)

src/lib.rs
```diff
@@ -1070,7 +1097,7 @@ pub unsafe extern "C" fn wgpuCommandEncoderBeginRenderPass(
                 .collect(),
         ),
         depth_stencil_attachment: depth_stencil_attachment.as_ref(),
-        timestamp_writes: None,
+        timestamp_writes: timestamp_writes.as_ref(),
         occlusion_query_set: None,
```
rajveermalviya: Should have said this in #299, but can you add a TODO comment here for populating `occlusion_query_set ` via an ext.

rajveermalviya (changes requested):
Thanks! Looks good overall!

Just a nit, but pushing through

rajveermalviya (approved):

PR #299, opened 2023-09-30
Add option for gles3 minor version in instance extras

eliemichel:
This was added to `wgpu-core` lately (commit extracted from #296 )


rajveermalviya (approved):
Thanks, Looks good!
But let's use latest trunk of wgpu, it should fix the CI failure.

eliemichel:
Done, thx for the suggestion!

rajveermalviya (approved):
Thanks!

PR #302, opened 2023-10-04
Limits fallback

### LINKED ISSUE #297: Adapters with lower-than-default limits
From the [3.6.2](https://www.w3.org/TR/webgpu/#limits) section of the spec about limits:

> Every adapter is guaranteed to support the default value or better.

But it turns out that in practice `wgpu-native` may create adapters that have more constrained limits.

For instance, somebody reported to me that they have `max_storage_textures_per_shader_stage` set to 2 in the supported limits, while spec says it must be at least 4. As a consequence, this adapter is not able to create any device using the default limits:

```
Failed to get a WebGPU device: Validation Error

Caused by:
    Limit 'max_storage_textures_per_shader_stage' value 4 is better than allowed 2
```

I see two reasonable ways to address this:

 - **(A)** Stick to WebGPU's spec, and do not create an adapter if it does not support the default limits.
 - **(B)** When device limits are not specified (set to nullptr), use the min of spec's default value and adapter's supported value as the default spec.

Solution **(B)** feels to me like a good trade off, as it does not obsoletes devices that are too limited, but still enables one to ignore the question of limits by setting them to `nullptr` in the device descriptor.

edit: [This quote](https://github.com/gpuweb/gpuweb/issues/1069) confirms that **(A)** isn't necessary thus **(B)** sounds better:

> native, non-web-exposed WebGPU implementations aren't beholden any minimum requirements here, or to the standard. So they may support older or less capable devices with feature emulation or compatibility gaps compared to the official spec.

rajveermalviya:
Fixes #297

#### Context
Support creating devices with lower than [`Limits::default()`](https://docs.rs/wgpu-types/0.17.0/src/wgpu_types/lib.rs.html#926-960) limits, by falling back to [`Limits::downlevel_default()`](https://docs.rs/wgpu-types/0.17.0/src/wgpu_types/lib.rs.html#964-996) if adapter's limits are lower than [`Limits::default()`](https://docs.rs/wgpu-types/0.17.0/src/wgpu_types/lib.rs.html#926-960). (checked by using [`wgt::Limits::check_limits()`](https://docs.rs/wgpu-types/latest/wgpu_types/struct.Limits.html#method.check_limits))

#### Testing
untested, but added a unit test

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -1,5 +1,3 @@
-use wgc::Label;
-
 use crate::native;
 use crate::utils::{make_slice, ptr_into_label, ptr_into_pathbuf};
 use crate::{follow_chain, map_enum};
@@ -298,12 +296,15 @@ pub fn map_instance_descriptor(
 pub fn map_device_descriptor<'a>(
     des: &native::WGPUDeviceDescriptor,
     extras: Option<&native::WGPUDeviceExtras>,
-) -> (wgt::DeviceDescriptor<Label<'a>>, *const std::ffi::c_char) {
+) -> (
+    wgt::DeviceDescriptor<wgc::Label<'a>>,
+    *const std::ffi::c_char,
+) {
     let limits = unsafe { des.requiredLimits.as_ref() }.map_or(
         wgt::Limits::default(),
         |required_limits| unsafe {
             follow_chain!(
-                map_required_limits(required_limits,
+                map_required_limits((required_limits),
                 WGPUSType_RequiredLimitsExtras => native::WGPURequiredLimitsExtras)
             )
         },
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -295,16 +295,20 @@ pub fn map_instance_descriptor(
 #[inline]
 pub fn map_device_descriptor<'a>(
     des: &native::WGPUDeviceDescriptor,
+    use_downlevel: bool,
     extras: Option<&native::WGPUDeviceExtras>,
 ) -> (
     wgt::DeviceDescriptor<wgc::Label<'a>>,
     *const std::ffi::c_char,
 ) {
     let limits = unsafe { des.requiredLimits.as_ref() }.map_or(
-        wgt::Limits::default(),
+        match use_downlevel {
+            true => wgt::Limits::downlevel_defaults(),
+            false => wgt::Limits::default(),
+        },
         |required_limits| unsafe {
             follow_chain!(
-                map_required_limits((required_limits),
+                map_required_limits((required_limits, use_downlevel),
                 WGPUSType_RequiredLimitsExtras => native::WGPURequiredLimitsExtras)
             )
         },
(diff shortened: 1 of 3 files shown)


almarklein (approved):
LGTM, but I'm not fully grasping how this works in practice.

IIUC, when you have an adapter that supports the default limits, it will create a device with these limits (unless the user specifies specific limits). If there is (at least) one default limit that the adapter cannot provide, the default limits (all of 'm) fall back to a lower standard (GLES3 compatible-ish).

rajveermalviya:
> IIUC, when you have an adapter that supports the default limits, it will create a device with these limits (unless the user specifies specific limits). If there is (at least) one default limit that the adapter cannot provide, the default limits (all of 'm) fall back to a lower standard (GLES3 compatible-ish).

Correct, also do note that these default limits are only used in following conditions:
- got `NULL` for descriptor in `wgpuAdapterRequestDevice`
- got `NULL` for `WGPUDeviceDescriptor.requiredLimits`
- used as base for limits with value of [`WGPU_LIMIT_[U32/U64]_UNDEFINED`](https://github.com/webgpu-native/webgpu-headers/blob/d12388575785f254b5c186667e0c1b79e4da46b1/webgpu.h#L72-L73) for `WGPUDeviceDescriptor.requiredLimits`

If user is relying on these defaults, then with this behaviour user would have to only worry about two possible levels of device types, one with default webgpu limits another with downlevel limits. If we do something else like try to use the webgpu default limits and clamping to available adapter limits - then it would no longer be the case.

A more critical user should not rely on these defaults, they should do their own investigation of possible device limits for their app/game and pass them for device creation.

rajveermalviya:
Another option is to use whatever adapter limits we get as the default for device creation, and tell the user that they should pass their own limits if they want anything else.

rajveermalviya:
> Another option is to use whatever adapter limits we get as the default for device creation, and tell the user that they should pass their own limits if they want anything else.

https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/vulkan/DeviceVk.cpp;l=106-108;drc=f5bdc89c7395ed24f1b8d196a3bdd6232d5bf771

~Seems like this is what dawn is doing currently.~ Maybe lets do that here too? WDYT @almarklein

almarklein:
So what happens when the user provides values for some limits, and `WGPU_LIMIT_[U32/U64]_UNDEFINED` for some others, and for one of these unspecified limits the adapter is not able to provide the webgpu defaults. Would it fall back to the downlevel limit (for that limit only) or fail to create the device?

This PR can be merged regardless of the answer, as far as I am concerned 👍

rajveermalviya:
Currently, if user provides some limit values and provides undefined for some (`WGPU_LIMIT_[U32/U64]_UNDEFINED`), then wgpu-native will use webgpu default limits for the undefined ones, and in case even one of the adapter supported limit is lower than the corresponding webgpu default limit then wgpu-native will use downlevel limits for all those undefined ones instead of webgpu default limits. For the non-undefined ones the values provided will be used as-is.

rajveermalviya:
And in case of limits that are exposed by wgpu-core are not present in `webgpu.h` we provide those extra limits in `wgpu.h`. If user doesn't provide the ext in `nextInChain` then wgpu-native will use whatever default/downlevel_default value for those limits.

almarklein:
Thanks for the explanation! All clear now 😄

PR #308, opened 2023-10-28
ci: setup cache

rajveermalviya:


almarklein (approved):
There's some formatting so its a bit hard what code actually changed, but I think you only added these blocks, right?
```yml
- name: Setup caching
        uses: Swatinem/rust-cache@v2
        with:
          key: build-${{ env.TARGET }}-${{ env.CACHE_SUFFIX }}
```
And IIUC they cache (versioned) Rust dependencies, so the builds are faster.

rajveermalviya:
Yeah, formatting is done via the Prettier extension from vscode. So thought I had to include it too, to make the review easier I separated them in 2 different commits.

Thanks to this caching, it reduces the workflow times significantly: [after caching](https://github.com/gfx-rs/wgpu-native/actions/runs/6677270012/usage), [before caching](https://github.com/gfx-rs/wgpu-native/actions/runs/6677006809/usage)

almarklein:
>  to make the review easier I separated them in 2 different commits.

Ah! Nice thinking, I missed that initially, but very clear indeed 👍 

Nice speedup!

PR #315, opened 2023-11-09
fix some samples releasing a null surface texture

radgeRayden:


rajveermalviya (approved):

almarklein:
Out of curiosity, what happened before this change? Did it lead to a panic?

rajveermalviya:
~In theory this should just be a sanity check (we have those in the example all over the place) because our impl of `wgpuSurfaceGetCurrentTexture` never returns `NULL` for texture.~ --- this statement is incorrect

rajveermalviya:
Looking through the implementation of `wgpuSurfaceGetCurrentTexture` & in wgpu-core, library will return null for all of these three cases (timeout, outdated, lost). So, the texture release there is not needed at all. Or we can conform to be consistent with all other functions which never return null for a wgpu object.

radgeRayden:
> Out of curiosity, what happened before this change? Did it lead to a panic?

In the C samples resizing the window would cause the application to hang. On my non C application for whatever reason I got the proper panic.

almarklein:
> In the C samples resizing the window would cause the application to hang. On my non C application for whatever reason I got the proper panic.

Hey, this sounds a lot like https://github.com/pygfx/wgpu-py/issues/352 cc @Korijn

PR #320, opened 2023-11-22
GetBindGroupLayout 'raises error' instead of panicing on invalid index

almarklein:
Ref discussion in https://github.com/pygfx/wgpu-py/pull/423

We have `wgpuRenderPipelineGetBindGroupLayout` and `wgpuComputePipelineGetBindGroupLayout`. End-users calling this code (e.g. people writing code against wgpu-py), may accidentally use an invalid index, which now causes a panic. In order for a wrapper to handle such errors gracefully, they must either know the number of layouts, or the call should not panic but generate an error instead. This PR does the latter (leading to a Python exception in wgpu-py).

Tested with wgpu-py (on the v0.18.1.1 tag).


almarklein:
~@rajveermalviya  if this looks good I'll also add it for the compute pipeline.~ done

rajveermalviya (approved):
Thanks!

rajveermalviya:
Will create a backport to `v0.18.1.x` later.

PR #322, opened 2023-11-23
Add missing webgpu features

almarklein:
The features specified by WebGPU (copied from https://gpuweb.github.io/gpuweb/#gpufeaturename):
```idl
enum GPUFeatureName {
    "depth-clip-control",
    "depth32float-stencil8",
    "texture-compression-bc",
    "texture-compression-etc2",
    "texture-compression-astc",
    "timestamp-query",
    "indirect-first-instance",
    "shader-f16",
    "rg11b10ufloat-renderable",
    "bgra8unorm-storage",
    "float32-filterable",
};
```

Features by wgpy-core: https://docs.rs/wgpu-types/0.18.0/wgpu_types/struct.Features.html

This PR adds one missing feature: "bgra8unorm-storage". (The "timestamp-query" was just moved so the order of the list is the same as in the spec.)

The final missing feature "float32-filterable" I cannot add because wgpu-core does not define it yet. 

We are interested in having it, since we rely on it in pygfx because it's somewhat crucial in e.g. volume rendering. So I looked a bit into wgpu-core, and found [the pr that adds bgra8unorm-storage](https://github.com/gfx-rs/wgpu/pull/4228). Seems like quite a bit of work, so for now I'm going to leave it at leaving the code for this feature commented here. We can keep using `TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES` (which includes float32-filterable)  for now. 😅

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -1059,9 +1059,6 @@ pub fn features_to_native(features: wgt::Features) -> Vec<native::WGPUFeatureNam
     if features.contains(wgt::Features::DEPTH32FLOAT_STENCIL8) {
         temp.push(native::WGPUFeatureName_Depth32FloatStencil8);
     }
-    if features.contains(wgt::Features::TIMESTAMP_QUERY) {
-        temp.push(native::WGPUFeatureName_TimestampQuery);
-    }
     if features.contains(wgt::Features::TEXTURE_COMPRESSION_BC) {
         temp.push(native::WGPUFeatureName_TextureCompressionBC);
     }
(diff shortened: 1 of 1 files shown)


almarklein:
I wonder 🤔  if we can/should fake float32-filterable by enabling `TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES` instead?

cwfitzgerald:
FYI bgra8 storage was really hard because it was the first to need to use a special spirv tag for storage textures.

Flost32 filterable should just be some logic checks

almarklein:
> Flost32 filterable should just be some logic checks

Ok, you nerd-sniped me now.

almarklein:
https://github.com/gfx-rs/wgpu/pull/4759

Issue #142, opened 2021-10-04
Build binaries for linux-arm64 (e.g. raspberry pi)


almarklein:
Should be easy to add now.

PR #326, opened 2023-11-29
add linux/aarch64 support.

dskkato:
This PR adds configurations to create binary files for arm64 architecture, which are necessary for platforms like Raspberry Pi. The build process takes approximately 30-40 minutes due to the use of QEMU. Currently, I am testing the functionality using Raspberry Pi 4 and wgpu-py, and so far, it seems that the library loading is successful.

You can use the artifacts in the CD from the following link. Feel free to use it for testing as needed.

https://github.com/dskkato/wgpu-native/releases/tag/v0.18.1.3

ss_raspi

Verified environments:

```sh
dskkato@raspberrypi:~ $ uname -m
aarch64
dskkato@raspberrypi:~ $ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

Related to #142

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -73,13 +73,13 @@ jobs:
           name: dist
 
   # -----
-  linux-arm64:
+  linux-aarch64:
     # Config
-    name: release - linux-arm64
+    name: release - linux-aarch64
     runs-on: ubuntu-latest
     env:
       TARGET: aarch64-unknown-linux-gnu
-      ARCHIVE_NAME: wgpu-linux-arm64
+      ARCHIVE_NAME: wgpu-linux-aarch64
       TOOLCHAIN: stable-aarch64-unknown-linux-gnu
       IMAGE: manylinux_2_28_aarch64
     steps:
@@ -279,7 +279,7 @@ jobs:
   publish:
     name: Publish Github release
     needs:
-      [linux-x86_64, linux-arm64, windows-x86_64, windows-i686, macos-x86_64, macos-arm64]
+      [linux-x86_64, linux-aarch64, windows-x86_64, windows-i686, macos-x86_64, macos-arm64]
     runs-on: ubuntu-latest
     if: success() && contains(github.ref, 'tags/v')
     steps:
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -48,10 +48,6 @@ jobs:
       - name: Set WGPU_NATIVE_VERSION
         run: echo "WGPU_NATIVE_VERSION=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
         shell: bash
-      - name: Setup caching
-        uses: Swatinem/rust-cache@v2
-        with:
-          key: build-${{ env.TARGET }}-${{ env.CACHE_SUFFIX }}
       # Build
       - name: Build
         run: |
@@ -94,10 +90,6 @@ jobs:
       - name: Set WGPU_NATIVE_VERSION
         run: echo "WGPU_NATIVE_VERSION=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
         shell: bash
-      - name: Setup caching
-        uses: Swatinem/rust-cache@v2
-        with:
-          key: build-${{ env.TARGET }}-${{ env.CACHE_SUFFIX }}
       # prepare qemu for cross-compilation
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -48,10 +48,65 @@ jobs:
       - name: Set WGPU_NATIVE_VERSION
         run: echo "WGPU_NATIVE_VERSION=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
         shell: bash
+      - name: Setup caching
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: build-${{ env.TARGET }}-${{ env.CACHE_SUFFIX }}
       # Build
       - name: Build
         run: |
           CID=$(docker create -t -w /tmp/wgpu-native -v $PWD:/tmp/src:ro -e TARGET -e ARCHIVE_NAME -e WGPU_NATIVE_VERSION quay.io/pypa/$IMAGE bash -c "\
+            cp -r /tmp/src/. . && \
+            rm -rf ./dist && \
+            export PATH=/root/.cargo/bin:\$PATH && \
+            export USER=root && \
+            curl --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain none && \
+            rustup toolchain install --no-self-update $TOOLCHAIN && \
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -48,6 +48,10 @@ jobs:
       - name: Set WGPU_NATIVE_VERSION
         run: echo "WGPU_NATIVE_VERSION=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
         shell: bash
+      - name: Setup caching
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: build-${{ env.TARGET }}-${{ env.CACHE_SUFFIX }}
```
rajveermalviya: AFAIK, the docker container can't benefit from this action - we would have to implement something else for that.
So, for now it's fine to leave caching out for the container builds.

.github/workflows/cd.yml
```diff
@@ -72,6 +76,57 @@ jobs:
           path: dist
           name: dist
 
+  # -----
+  linux-arm64:
```
rajveermalviya:
```suggestion
  linux-aarch64:
```
Let's use same notation as Rust/LLVM, here and anywhere below.

.github/workflows/cd.yml
```diff
@@ -72,6 +76,57 @@ jobs:
...
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: Set WGPU_NATIVE_VERSION
+        run: echo "WGPU_NATIVE_VERSION=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+        shell: bash
+      - name: Setup caching
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: build-${{ env.TARGET }}-${{ env.CACHE_SUFFIX }}
```
rajveermalviya: same for caching here as above.

rajveermalviya (changes requested):
Nice, Thanks!

almarklein:
Very nice! In #148 I did an earlier attempt using `armv7-unknown-linux-gnueabihf` as a compilation target. Do you know anything about what this difference means in practice? IIRC Raspberry Pi 3 required arm7, perhaps version 4 does not?

.github/workflows/cd.yml
dskkato: I also considered using "aarch64," but for macOS, it is "arm64". Is it OK to have different naming rules for Linux and macOS? Personally, I prefer to use "aarch64" for macOS as well.

.github/workflows/cd.yml
dskkato: I just tried to see if there was a way to reduce the time it was taking to build on QEMU. I will revert them.

.github/workflows/cd.yml
dskkato: removed.

.github/workflows/cd.yml
dskkato: removed.

.github/workflows/cd.yml
dskkato: Changed arm64 to aarch64 on linux side only.

rajveermalviya: Hmm I hadn't checked the macOS build but indeed that's a problem, we need to keep the archive arch names consistent for different OSes. 
But we are already not consistent with it, it seems:
| Go | Rust/LLVM |
| - | - |
| `amd64` | `x86_64` |
| `386` | `i686` | 
| `arm64` | `aarch64` |

I preferred `aarch64` because we are already using the LLVM notation for `x86_64`, so I would prefer the macOS archive & this new Linux archive to use same naming. But that would break existing scripts/build systems that fetch macOS archives.

@almarklein, what does Python build system prefer, and WDYT about changing macOS archive to use `aarch64` instead of current `arm64`.

dskkato:
@almarklein 

> I did an earlier attempt using `armv7-unknown-linux-gnueabihf` as a compilation target. Do you know anything about what this difference means in practice? IIRC Raspberry Pi 3 required arm7, perhaps version 4 does not?

Basically, "armv7" is for Raspberry Pi OS (32-bit), while "aarch64" is for Raspberry Pi OS (64-bit). The 64-bit OS for Raspberry Pi has been available for about two years and can be used with Raspberry Pi 3B models and newer. 

https://www.raspberrypi.com/news/raspberry-pi-os-64-bit/

Since it is built for the aarch64 architecture, it may not run on 32-bit OS environments on the same hardware. I've added information about my verification results to the PR description.

I also attempted cross-compiling for armv7, but I decided to skip it this time because the latest manylinux only supports aarch64 (manylinux2014_armv7l seems to be the last one). While it is possible to build for armv7 by downloaing compilers from somewhere (ARM, linaro, ...) for cross-compile, the generalizability of the build results is quite uncertain in that case.

.github/workflows/cd.yml
almarklein: This only really affects the name of the zipfile to download from the release page.

We [construct the name of the zipfile](https://github.com/pygfx/wgpu-py/blob/4478b19d03598036169eaeed46bd05c4ac0dcf29/download-wgpu-native.py#L79-L101) in wgpu-py, and we can easily update this as long as there's some clear logic to follow. 

As far as how Python names it's binary wheels (e.g. [for numpy](https://pypi.org/project/numpy/#files)):
* For Windows it's `win_amd64`  or `win32` (not sure what an arm build would be called).
* For Linux it's `x86_64` or `aarch64`.
* For MacOS it's `x86_64`  or `arm64`.

I'd decide whatever makes the most sense from the pov of C/Rust developers.

rajveermalviya: > This only really affects the name of the zipfile to download from the release page.

Yes, also not so much of a problem for users with versioned dependency loading/downloading (which is probably everyone).

> For Windows it's win_amd64 or win32, For Linux it's x86_64 or aarch64, For MacOS it's x86_64 or arm64.

Wild, each OS uses different arch notation. I would prefer it to be consistent though.

So, let's update the macOS build to use `aarch64` too. @dskkato can you please include that too in this PR?

dskkato: I've changed macOS `arm64` to `aarch64`. Please check the results in the CD pipeline on my forked repository.

https://github.com/dskkato/wgpu-native/actions/runs/7041779247

rajveermalviya (approved):

almarklein:
Thanks a lot!

dskkato:
Could we please have a release of the version after this patch was merged? I'd like to use it for validation in the Raspberry Pi environment.

almarklein:
> Could we please have a release of the version after this patch was merged?

I think we can backport this to the v0.18 branch and tag v0.18.1.3, what do you think @rajveermalviya ?

rajveermalviya:
Created release tag: https://github.com/gfx-rs/wgpu-native/compare/v0.18.1.2...v0.18.1.3

PR #333, opened 2023-12-15
Build packages for Mac Catalyst

ifiddynine:
We utilize Mac Catalyst heavily internally, and would love to see prebuilt binaries for that platform on top of MacOS.

Note: Untested on the wgpu-native build machines since I'm not able to.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -275,6 +275,78 @@ jobs:
           path: dist
           name: dist
 
+  # -----
+  maccatalyst-x86_64:
+    # Config
+    name: release - maccatalyst-x86_64
+    runs-on: macos-latest
+    env:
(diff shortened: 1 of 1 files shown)


almarklein:
I'm not 100% sure, but I think that on your fork you can manually trigger the CD build. That should build the new binaries so they can be tested.

almarklein:
The cd script was refactored recently. The builds are now grouped by OS, using a matrix to build various flavours of e.g. Mac builds. Should be fairly easy to adopt.

But would be great if you could first test the builds, by triggering the CD on your own fork.

almarklein:
Closing, since this is over a year old. Feel free to re-open if you want to revive this.

There's two things holding this back:
* The cd script has been refactored, so it will have to be adjusted accordingly.
* Get the builds via CD on ones own fork, so the build can be tested/verified.

PR #349, opened 2024-01-16
avoid returning null strings

rajveermalviya:
Avoid returning null pointer for message strings, instead allocate an empty string, it creates undefined behavior if caller doesn't check message pointer and streams it to `cout` directly.


almarklein (approved):
Nice!

PR #353, opened 2024-01-21
Simplify examples

rajveermalviya:
Remove the unnecessary `ASSERT_CHECK` contraption I'd made to gracefully exit instead of an `assert`.
Now examples are much more readable without the gotos/labels & forward declarations, also resulting in lesser lines of code!


almarklein (approved):
Nice!

PR #357, opened 2024-01-24
workaround queue references on single QueueId

### LINKED ISSUE #356: Releasing any WGPUQueue destroys all WGPUQueues
E.g., just doing this:
```c
{
    WGPUQueue temp = wgpuDeviceGetQueue(device);
    wgpuQueueRelease(temp);
}
```
will cause usage of any other queue from that device to fail with something like
```
thread '<unnamed>' panicked at C:\Users\runneradmin\.cargo\git\checkouts\wgpu-53e70f8674b08dd4\b8a8ff6\wgpu-core\src\storage.rs:113:39:
Queue[Id(0,1,vk)] does not exist
```

I suspect this is related/the same underlying issue as https://github.com/gfx-rs/wgpu/issues/5092

Reproducer:
```c
#include "webgpu.h"
#include "wgpu.h"
#include <assert.h>
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <stdint.h>

typedef struct Demo {
  WGPUInstance instance;
  WGPUAdapter adapter;
  WGPUDevice device;
} Demo;

static void handle_request_adapter(WGPURequestAdapterStatus status,
                                   WGPUAdapter adapter, char const *message,
                                   void *userdata) {
  if (status == WGPURequestAdapterStatus_Success) {
    Demo* demo = userdata;
    demo->adapter = adapter;
  } else {
    printf("request_adapter status=%#.8x message=%s\n", status, message);
  }
}

static void handle_request_device(WGPURequestDeviceStatus status,
                                  WGPUDevice device, char const *message,
                                  void *userdata) {
  if (status == WGPURequestDeviceStatus_Success) {
    Demo* demo = userdata;
    demo->device = device;
  } else {
    printf("request_device status=%#.8x message=%s\n", status, message);
  }
}

int main(int argc, char *argv[]) {
  Demo demo = {0};
  demo.instance = wgpuCreateInstance(NULL);
  assert(demo.instance);

  WGPURequestAdapterOptions options = {0};
  wgpuInstanceRequestAdapter(demo.instance, &options, handle_request_adapter, &demo);
  assert(demo.adapter);

  wgpuAdapterRequestDevice(demo.adapter, NULL, handle_request_device, &demo);
  assert(demo.device);

  WGPUQueue queue = wgpuDeviceGetQueue(demo.device);
  assert(queue);

  if(argc <= 1) {
    printf("Only getting one queue; this should complete normally!\n");
  } else {
    printf("Requesting & releasing another queue!\n");
    WGPUQueue queue2 = wgpuDeviceGetQueue(demo.device);
    wgpuQueueRelease(queue2);
  }

  WGPUBufferDescriptor buff_desc = {.usage = WGPUBufferUsage_CopyDst, .size = 256};
  WGPUBuffer buff = wgpuDeviceCreateBuffer(demo.device, &buff_desc);
  uint8_t data[256] = {0};
  wgpuQueueWriteBuffer(queue, buff, 0, data, 256);

  printf("Done without panic!\n");

  return 0;
}
```

rajveermalviya:
Previously, we created a `WGPUQueueImpl` from the same `queue_id` that was created when requesting device, this behavior resulted in a bug where any call to `wgpuQueueRelease` would drop the `queue_id`, making the ref-counting useless.

This PR fixes it by instead storing an `Arc` of `QueueId` in `WGPUQueueImpl`, allowing us to correctly provide the ref-counting over the single `queue_id` we get during device initialization.

Fixes #356 

#### Testing
Ran the repro code in the mentioned issue.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -171,10 +171,23 @@ impl Drop for WGPUComputePipelineImpl {
     }
 }
 
+struct QueueId {
+    context: Arc<Context>,
+    id: id::QueueId,
+}
+impl Drop for QueueId {
+    fn drop(&mut self) {
+        if !thread::panicking() {
+            let context = &self.context;
+            gfx_select!(self.id => context.queue_drop(self.id));
+        }
+    }
+}
+
 pub struct WGPUDeviceImpl {
     context: Arc<Context>,
     id: id::DeviceId,
-    queue_id: id::QueueId,
+    queue: Arc<QueueId>,
     error_sink: ErrorSink,
 }
 impl Drop for WGPUDeviceImpl {
@@ -228,17 +241,8 @@ impl Drop for WGPUQuerySetImpl {
 }
 
 pub struct WGPUQueueImpl {
-    context: Arc<Context>,
-    id: id::QueueId,
+    queue: Arc<QueueId>,
     error_sink: ErrorSink,
-}
-impl Drop for WGPUQueueImpl {
-    fn drop(&mut self) {
-        if !thread::panicking() {
-            let context = &self.context;
-            gfx_select!(self.id => context.queue_drop(self.id));
-        }
-    }
 }
 
 pub struct WGPURenderBundleImpl {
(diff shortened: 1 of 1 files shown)


almarklein (approved):
Nice one!

PR #358, opened 2024-01-27
examples: Don't build docs for glfw.

waywardmonkeys:
Also, update to using `ON` rather than `1` for the flag for using Wayland.


rajveermalviya (approved):
Nice, Thanks!

PR #359, opened 2024-01-27
wgpu.h: Don't require preprocessing to use.

waywardmonkeys:
Right now, `wgpu.h` can be included directly and it then includes `webgpu-headers/webgpu.h`. This is altered during creation of a distribution package to remove the `"webgpu-headers"`, but by adding `ffi/webgpu-headers` to the include search path, we can avoid having to preprocess the `wgpu.h`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -42,8 +42,8 @@ fn main() {
         ("WGPUTextureView", "WGPUTextureViewImpl"),
     ];
     let mut builder = bindgen::Builder::default()
-        .header("ffi/webgpu-headers/webgpu.h")
         .header("ffi/wgpu.h")
+        .clang_arg("-Iffi/webgpu-headers")
         .parse_callbacks(Box::new(bindgen::CargoCallbacks::new()))
         .blocklist_function("wgpuGetProcAddress")
         .prepend_enum_name(false)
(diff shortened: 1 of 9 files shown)


waywardmonkeys:
I ran across this while working on some cmake stuff regarding #251 

cc: @eliemichel

build.rs
```diff
@@ -44,6 +44,7 @@ fn main() {
     let mut builder = bindgen::Builder::default()
         .header("ffi/webgpu-headers/webgpu.h")
         .header("ffi/wgpu.h")
```
rajveermalviya: Don't need to specify it here if including via include path.

```suggestion
    let mut builder = bindgen::Builder::default()
        .header("ffi/wgpu.h")
```

rajveermalviya (approved):
Great work!
Apart from one nit LGTM.

build.rs
waywardmonkeys: Out at dinner at the moment, will try this in 90 minutes or so.

waywardmonkeys: Okay, removed the include of `ffi/webgpu-headers/webgpu.h` and confirmed that the generated `bindings.rs` remains the same.

waywardmonkeys:
@rajveermalviya Hopefully good to go now!

rajveermalviya (approved):
Thanks! LGTM

Verified locally that the generated package remains the same.

waywardmonkeys:
Thanks… another Pr coming up soon!

PR #360, opened 2024-01-27
examples: Improve glfw build integration.

waywardmonkeys:
Use the `glfw` target from cmake to configure other build targets.


waywardmonkeys:
The failed clippy iOS check looks like a GitHub Actions error, not from my changes.

rajveermalviya (approved):

PR #361, opened 2024-01-27
bindgen: Only bind `WGPU.*` and `wgpu.*`.

waywardmonkeys:
The bindings only need to include things from the WGPU API, not stuff from the headers that they happen to include from the compiler and libc.


rajveermalviya (approved):
Thanks!

I checked the generated `bindings.rs` and it's much smaller, also bindgen runs a little bit faster.

PR #379, opened 2024-04-04
Fix device limits diverging from base limits by default

### LINKED ISSUE #5490: Device limits reports a max texture size that is bigger than defaults even if not requested
**Description**
After creating the device without asking for any particular limits, I get a maxTextureDimension2D value that is way larger than the default. On my machine with an RTX3080 I get a limit of 32768. Other limits might be exceeding defaults, but I haven't tested many of them.

**Repro steps**
It is observable by slightly modifying the hello_triangle example to print the limits:
```diff
diff --git a/examples/src/hello_triangle/mod.rs b/examples/src/hello_triangle/mod.rs
index faa1db8f8..350c03f8d 100644
--- a/examples/src/hello_triangle/mod.rs
+++ b/examples/src/hello_triangle/mod.rs
@@ -38,6 +38,9 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
         .await
         .expect("Failed to create device");
 
+    let limits = device.limits();
+    print!("{0}", limits.max_texture_dimension_2d);
+
     // Load the shaders from disk
     let shader = device.create_shader_module(wgpu::ShaderModuleDescriptor {
         label: None,
```

**Expected vs observed behavior**
I expected the device to report a max texture size of 8196 as per [the spec](https://gpuweb.github.io/gpuweb/#limits).

**Platform**
Running on Arch Linux default kernel, NVIDIA RTX3080 driver version 550.67. I have observed the problem in the wgpu sample and on my own application running on top of wgpu-native.

radgeRayden:
Fixes https://github.com/gfx-rs/wgpu/issues/5490. It was pointed out to me that these lines were exactly the same as the ones in the rust demo, which smells like a copy paste mistake.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -794,7 +794,26 @@ pub unsafe extern "C" fn wgpuAdapterRequestDevice(
             return;
         }
     };
-    let base_limits = get_base_device_limits_from_adapter_limits(&adapter_limits);
+
+    let default_limits = wgt::Limits::default();
+    let dim_1d = std::cmp::min(
+        adapter_limits.max_texture_dimension_1d,
+        default_limits.max_texture_dimension_1d,
+    );
+    let dim_2d = std::cmp::min(
+        adapter_limits.max_texture_dimension_2d,
+        default_limits.max_texture_dimension_2d,
+    );
+    let dim_3d = std::cmp::min(
+        adapter_limits.max_texture_dimension_2d,
+        default_limits.max_texture_dimension_3d,
+    );
+    let base_limits = wgt::Limits {
+        max_texture_dimension_1d: dim_1d,
+        max_texture_dimension_2d: dim_2d,
+        max_texture_dimension_3d: dim_3d,
+        ..get_base_device_limits_from_adapter_limits(&adapter_limits)
+    };
 
     let (desc, trace_str, device_lost_handler) = match descriptor {
         Some(descriptor) => {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -794,26 +794,7 @@ pub unsafe extern "C" fn wgpuAdapterRequestDevice(
             return;
         }
     };
-
-    let default_limits = wgt::Limits::default();
-    let dim_1d = std::cmp::min(
-        adapter_limits.max_texture_dimension_1d,
-        default_limits.max_texture_dimension_1d,
-    );
-    let dim_2d = std::cmp::min(
-        adapter_limits.max_texture_dimension_2d,
-        default_limits.max_texture_dimension_2d,
-    );
-    let dim_3d = std::cmp::min(
-        adapter_limits.max_texture_dimension_2d,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -794,10 +794,7 @@ pub unsafe extern "C" fn wgpuAdapterRequestDevice(
             return;
         }
     };
-    let base_limits = get_base_device_limits_from_adapter_limits(&adapter_limits)
-        // make sure we use the texture resolution limits from the adapter,
-        // so we can support images the size of the surface.
-        .using_resolution(adapter_limits);
```
almarklein, line 800: Should we still do *something* to at least support the surface size?

radgeRayden: > Should we still do _something_ to at least support the surface size?

The default limit is 8196, which seems like "good enough everywhere" for surfaces. The only reason you'd be using a lesser limit would be with downlevels, and then you simply can't support that big a texture anyway. This is however a terrifying thought so I'm curious as well.

radgeRayden: PS: I realize now the purpose of that line might be to make the size _smaller_ if necessary. In that case the code should choose the smaller default. But imo that should be handled by the base_limits function.

rajveermalviya: > I realize now the purpose of that line might be to make the size smaller if necessary

Lets keep it per spec for now.

rajveermalviya (dismissed):

src/lib.rs
cwfitzgerald: The reason that's there in the wgpu examples is because we use the downlevel limits, which only requires tex size of 2k

rajveermalviya: Now that I remember that's the same reason it was included here too, we determine _some_ "default" limits when user passes nothing, and it can be either of `downlevel_webgl2`/`downlevel`/`default.
So, when downlevel or lower is selected we need to use maximum texture size supported by the adapter, so that it can support display sizes more than 2K if the adapter supports it.

radgeRayden: The issue here (as stated in the issue) is that this gets applied unconditionally, so by default I'm getting limits much larger than the spec mandates.

rajveermalviya: yeah, the fix here would be to clamp to webgpu spec default limits.

rajveermalviya (dismissed):

src/utils.rs
```diff
@@ -72,6 +74,29 @@ pub fn get_base_device_limits_from_adapter_limits(adapter_limits: &wgt::Limits)
     }
     wgt::Limits::downlevel_webgl2_defaults()
 }
+#[inline]
+pub fn get_base_device_limits_from_adapter_limits(adapter_limits: &wgt::Limits) -> wgt::Limits {
```
radgeRayden, line 58: That's better, it really felt like the base_device_limits fn should handle that in the first place.

almarklein (approved):

PR #383, opened 2024-04-30
Update native headers to latest commit

PJB3005:
Everything is just marked as unimplemented for now, due to lacking support in main wgpu.

The biggest reason I did this is because the added depthSlice field in WGPURenderPassColorAttachment caused conflicts when trying to use up-to-date headers with wgpu-native.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -772,7 +772,7 @@ pub unsafe extern "C" fn wgpuAdapterHasFeature(
 pub unsafe extern "C" fn wgpuAdapterRequestDevice(
     adapter: native::WGPUAdapter,
     descriptor: Option<&native::WGPUDeviceDescriptor>,
-    callback: native::WGPURequestDeviceCallback,
+    callback: native::WGPUAdapterRequestDeviceCallback,
     userdata: *mut std::os::raw::c_void,
 ) {
     let (adapter_id, context) = {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -1172,6 +1172,10 @@ pub unsafe extern "C" fn wgpuCommandEncoderBeginRenderPass(
             make_slice(descriptor.colorAttachments, descriptor.colorAttachmentCount)
                 .iter()
                 .map(|color_attachment| {
+                    if color_attachment.depthSlice != 0 {
```
rajveermalviya:
```suggestion
if color_attachment.depthSlice != native::WGPU_DEPTH_SLICE_UNDEFINED {
                        log::warn!("Depth slice on color attachments is not implemented");
}
```
and then in examples update accordingly:

```c
(const WGPURenderPassColorAttachment){
    ...
    .depthSlice = WGPU_DEPTH_SLICE_UNDEFINED,
```

rajveermalviya (changes requested):
Apart from one nit, this looks good to me!
Thanks for working on this! :)

almarklein (approved):

src/lib.rs
```diff
@@ -1172,6 +1172,10 @@ pub unsafe extern "C" fn wgpuCommandEncoderBeginRenderPass(
             make_slice(descriptor.colorAttachments, descriptor.colorAttachmentCount)
                 .iter()
                 .map(|color_attachment| {
+                    if color_attachment.depthSlice != native::WGPU_DEPTH_SLICE_UNDEFINED {
+                        unimplemented!("Depth slice on color attachments is not implemented");
```
rajveermalviya: We can't panic by default because it will break pretty much all existing code, a `log::warn` would be better here like I mentioned in previous review.

rajveermalviya (approved):
Thanks!
I'll push a commit for this nit, and merge it now.

PR #387, opened 2024-05-23
Update to wgpu-core 0.20.0

almarklein:
Updating to https://github.com/gfx-rs/wgpu/releases/tag/v0.20.0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,9 +58,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.2.0"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1fdabc7756949593fe60f30ec81974b613357de856987752631dea1e3394c80"
+checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
 
 [[package]]
 name = "base64"
@@ -134,12 +134,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
 
 [[package]]
-name = "cc"
-version = "1.0.94"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "17f6e324229dc011159fcc089755d1e2e216a90d43a7dea6853ca740b84f35e7"
-
-[[package]]
 name = "cexpr"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


almarklein:
I'm attempting to update, but my I'm somewhat overwhelmed by the Rust type errors:

E.g. I get a ton of these; is the suggestion from the Rust compiler the way to go?
```
error[E0599]: no method named `as_mut` found for struct `Id` in the current scope
    --> src/lib.rs:4293:32
     |
4293 |     let encoder = pass.encoder.as_mut().unwrap();
     |                                ^^^^^^ method not found in `Id<RenderPassEncoder>`
     |
    ::: /Users/almar/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/convert/mod.rs:372:8
     |
372  |     fn as_mut(&mut self) -> &mut T;
     |        ------
     |        |
     |        the method is available for `Box<Id<RenderPassEncoder>>` here
     |        the method is available for `Box<&mut Id<RenderPassEncoder>>` here
     |        the method is available for `Box<&Id<RenderPassEncoder>>` here
     |
help: consider wrapping the receiver expression with the appropriate type
     |
4293 |     let encoder = Box::new(pass.encoder).as_mut().unwrap();
     |                   +++++++++            +
help: consider wrapping the receiver expression with the appropriate type
     |
4293 |     let encoder = Box::new(&mut pass.encoder).as_mut().unwrap();
     |                   +++++++++++++             +
help: consider wrapping the receiver expression with the appropriate type
     |
4293 |     let encoder = Box::new(&pass.encoder).as_mut().unwrap();
     |                   ++++++++++            +
```

rajveermalviya:
Yeah there are quite a lot of changes needed, [here's my branch](https://github.com/rajveermalviya/wgpu-native/commit/4b50d94479961ef9c626450338eebe5df69c2672) that I was working a while ago, it compiles and runs but has a ref counting race condition somewhere for the surface, so it segfaults at exit. I didn't get much time to work on finding it though.

rajveermalviya:
(Feel free to pull those changes in this PR :))

almarklein:
Ah, great to see you've already made the changes. That segfault sounds bad. I might also have a look.

almarklein:
I reviewed your changes, and it looks fine to me (albeit my Rust knowledge is limited). I especially checked the code related to surfaces, but cannot find anything suspicious of why the race condition was introduced. I also don't get a segfault on MacOS.

rajveermalviya:
This https://github.com/gfx-rs/wgpu/issues/5637 may be related, it's fix is added as a backport, so maybe will need to wait till next point release.
(I still need to verify by moving to wgpu trunk)

Update: It doesn't fix it unfortunately, but it's related to our `Arc` usage for Surface + Device.

ygdrasil-io:
Is it possible to publish a version even if we have some memory issue on it ? 
The issue could be track and fix on a futur minor version, so, peoples waiting for bugfixes/new features are not block.

rajveermalviya:
The segfault turned out to be fixed via https://github.com/gfx-rs/wgpu/pull/6052, #402 updates wgpu to `v22.1.0` which includes that fix and no longer segfaults.

PR #394, opened 2024-06-18
meson: install headers

natto1784:
Installing as a standalone library works due to `install: true` in `static_library()`. However, headers are not automatically installed. This commit adds the two headers available under `ffi/`.

Additionally, I have set `check: true` for the cargo command


almarklein (approved):
Looks fine by me ...

ffi/meson.build
```diff
@@ -0,0 +1,8 @@
+inc_dirs += [include_directories('.', is_system: true)]
+
+headers = files(
+  'wgpu.h',
+  'webgpu-headers/webgpu.h'
```
hmaarrfk, line 5: is it your intention for this package to "take over" the dependency of `webgpu.h`? Would it ever make sense to we `webgpu.h` alone without `wgpu-native`????

if so, installing `webgpu.h` might be strange.

almarklein: I'm a bit confused as to what you're asking. But I can try to explain how the two header files relate.

The `webgpu.h` is specified alongside `webgpu.idl` (the JS spec), and can be targeted by any implementation (wgpu-native and Dawn are two examples). That's why its in a separate repo. The `wgpu.h` contains additional functionality specific to desktop/wgpu. As far as I know, it is expected that both will stay separate for the foreseeable future.

Does this answer your question?

hmaarrfk: This was my understanding as well.

So to me, `webgpu.h` is not "owned" by the `wgpu-native` library and thus should not be installed by it.

Rather, some package integrator (this is REALLY annoying for C libraries since no single package manager has surfaced as a winner), is in charge of:

1. Installing the `webgpu.h` header.
2. Then starting the compiation of `wgpu-native`.

`wgpu-native` should be checking for compatibility with the installed header `webgpu.h` prior to starting.


So having `wgpu-native` copy over a vendoered version of `webgpu.h` is somewhat problematic long term.

In terms of how this is done in practice, for example, at conda-forge, i declare the build time dependency on `webgpu-headers`, but due to the strict version matching, I must also specify the version i want installed.

https://github.com/conda-forge/wgpu-native-feedstock/blob/main/recipe/meta.yaml#L27

Again, the large range of C package manager (mostly non-existant in a cross platform way without conda IMO) means that this is always done in an ad-hoc manner.

In either case, I think that the `webgpu.h` should not be installed here.

hmaarrfk:
(I don't meant to backseat drive this project, just noticed build related PRs when trying to check what changed "recently")

ffi/meson.build
almarklein: I'm still not sure if I follow :) Might be my limited understanding of C build systems ...

I'm not sure what you mean by wgpu-native *installing* webgpu.h.

The `webgpu-headers` repo is added as a git submodule, tagged to a specific commit. From time to time, the submodule is bumped to the latest version, and any necessary changes to the source of wgpu-native are made. I think you can consider (the specific version of) `webgpu.h` to be part of this repo. Not owned, but definitely pinned.

> So having wgpu-native copy over a vendoered version of webgpu.h is somewhat problematic long term.

How do you think this can become problematic?

hmaarrfk: let me try to see how this works in practice and followup with a larger issue if it is a problem.

thakns for your time!

natto1784: I think what @hmaarrfk is trying to say that webgpu-headers should be installed separately instead of installing its header with wgpu-native. I kind of agree with him but this is a relatively minor change I needed for my build system, feel free to change. Users will have to install webgpu-headers/webgpu.h separately though.

PR #395, opened 2024-06-21
Add 16bit norm and NV12 textures

davnotdev:
`Features::TEXTURE_FORMAT_16BIT_NORM` and `Features::TEXTURE_FORMAT_NV12` are currently exposed as native features. This PR adds a new enum `WGPUTextureFormatExtras` which will make the texture formats of the mentioned features accessible.

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -819,13 +819,15 @@ pub fn map_texture_format(value: native::WGPUTextureFormat) -> Option<wgt::Textu
         native::WGPUTextureFormat_ASTC12x10UnormSrgb => Some(wgt::TextureFormat::Astc { block: AstcBlock::B12x10, channel: AstcChannel::UnormSrgb }),
         native::WGPUTextureFormat_ASTC12x12Unorm => Some(wgt::TextureFormat::Astc { block: AstcBlock::B12x12, channel: AstcChannel::Unorm }),
         native::WGPUTextureFormat_ASTC12x12UnormSrgb => Some(wgt::TextureFormat::Astc { block: AstcBlock::B12x12, channel: AstcChannel::UnormSrgb }),
-        native::WGPUTextureFormatExtras_R16Unorm => Some(wgt::TextureFormat::R16Unorm),
-        native::WGPUTextureFormatExtras_R16Snorm => Some(wgt::TextureFormat::R16Snorm),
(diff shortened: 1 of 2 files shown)


src/conv.rs
```diff
@@ -819,6 +819,13 @@ pub fn map_texture_format(value: native::WGPUTextureFormat) -> Option<wgt::Textu
         native::WGPUTextureFormat_ASTC12x10UnormSrgb => Some(wgt::TextureFormat::Astc { block: AstcBlock::B12x10, channel: AstcChannel::UnormSrgb }),
         native::WGPUTextureFormat_ASTC12x12Unorm => Some(wgt::TextureFormat::Astc { block: AstcBlock::B12x12, channel: AstcChannel::Unorm }),
         native::WGPUTextureFormat_ASTC12x12UnormSrgb => Some(wgt::TextureFormat::Astc { block: AstcBlock::B12x12, channel: AstcChannel::UnormSrgb }),
+        native::WGPUTextureFormatExtras_R16Unorm => Some(wgt::TextureFormat::R16Unorm),
```
almarklein: Why not simply `WGPUTextureFormat_R16Unorm`?

almarklein: Oh, I guess because the prefix is the full enum name...

davnotdev: I'm a bit divided on the naming. I do believe in the importance of separating types that are outside the spec and following established precedent, but on the other hand, `WGPUTextureFormat_R16Unorm` feels much more natural and may become apart of the spec one day anyway. Let me which you would prefer.

almarklein: Right, from the point of view of wgpu-core the texture formats are equal, but from the pov of wgpu native they are not, so I'd prefer keeping the "extra" in the prefix. @rajveermalviya thoughts?

rajveermalviya: Yeah, there isn't much consistency in naming extended enums/structs in `wgpu.h`, we have enums like `WGPUNativeFeature` (extends `WGPUFeatureName`), and structs use the `Extras` suffix.

So I think we should have the `WGPUNative` prefix for extended enums here matching other extended enums, for consistency.

I'll make that change and then merge.

rajveermalviya (approved):
Thanks!

PR #402, opened 2024-08-04
Update wgpu to v22.1.0

rajveermalviya:
https://github.com/gfx-rs/wgpu/releases/tag/v22.1.0

#### Testing
All examples tested on macOS + Metal, no leaks, no segfaults. Including the new changes in error handling tested to be working correctly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,26 +23,22 @@ resolver = "2"
 [workspace.dependencies.wgc]
 package = "wgpu-core"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "5c5c8b1d4d2d965fbd10b290ee26f4e7eb158d7c"
-version = "22.0.0"
+tag = "v22.1.0"
 
 [workspace.dependencies.wgt]
 package = "wgpu-types"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "5c5c8b1d4d2d965fbd10b290ee26f4e7eb158d7c"
-version = "22.0.0"
+tag = "v22.1.0"
 
 [workspace.dependencies.hal]
 package = "wgpu-hal"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "5c5c8b1d4d2d965fbd10b290ee26f4e7eb158d7c"
-version = "22.0.0"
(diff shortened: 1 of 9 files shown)


yig:
What is blocking acceptance of this pull request?

Cargo.toml
```diff
@@ -23,20 +23,26 @@ resolver = "2"
 [workspace.dependencies.wgc]
 package = "wgpu-core"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "87576b72b37c6b78b41104eb25fc31893af94092"
-version = "0.19"
+rev = "5c5c8b1d4d2d965fbd10b290ee26f4e7eb158d7c"
+version = "22.0.0"
```
almarklein: Looks like it pins to 22.0.0, not 22.1.0?

rajveermalviya: Hmm, I'd used the values from wgpu's Cargo.toml:
https://github.com/gfx-rs/wgpu/blob/v22.1.0/Cargo.toml#L53-L70

But it looks like since we're using `git` dependencies, the actual version registered in `Cargo.lock` for `wgpu-core` (and friends) seems to be correct. I guess the best would be to not explicitly specify those `version` fields, and instead just let `cargo` determine them based on the git `tag`/`rev` fields.

For more info: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#the-role-of-the-version-key

almarklein: I like tagging to a tag, instead of a hash 😄

almarklein (approved):

rajveermalviya:
(The Android clippy CI failures is probably because Github Actions images dropped/updated the included Android NDK bundle — https://github.com/gfx-rs/wgpu-native/blob/91f1dc0cd270afddbe188f7444d75dee3cc562dc/.github/workflows/ci.yml#L88)

~Will create a separate PR to fix this.~ Opened #411

yig:
This is exciting! Will a tag and release be automatically generated?

rajveermalviya:
Released — https://github.com/gfx-rs/wgpu-native/releases/tag/v22.1.0.1

PR #404, opened 2024-08-04
Make it clear that windows builds use MSVC

Lisible:
I spent some time dealing with linking issues with MSYS2 UCRT64 before realizing that the auto-generated binaries for Windows are built using MSVC.

This patch adds that detail to the release note so that no one else has to figure that out.


rajveermalviya (approved):
Thanks!

PR #407, opened 2024-08-14
Add Jai bindings

SogoCZE:
Hello,

I've created a script in Jai for auto-generating bindings and building the libraries for Jai. So I thought it would make sense to list the bindings here.


cwfitzgerald (approved):
Thanks!

PR #409, opened 2024-08-15
Add perl bindings link to README.md

atrodo:
I have released a WebGPU perl module that uses wgpu-native, so I was hoping to add it to the list.


almarklein (approved):
Thanks!

PR #417, opened 2024-09-04
Change artifact folder structure

### LINKED ISSUE #410: Prebuilt releases header files location
For consuming the prebuilt releases from a build system (meson in my case), it would help if the headers were placed in a `[include]/webgpu` subfolder.

When building with other libraries than can use WebGPU (such as imgui) it seems expected to have WebGPU available in the form of `#include <webgpu/webgpu.h>` (note the leading directory).
However it is currently needed to manually move the headers in a `webgpu` subfolder after extracting the archive to achieve that and specify correct include paths in the build system.

almarklein:
Closes #410 

This updates the package command in the makefile to introduce a folder structure instead of putting all files in a flat namespace. The idea is to prepare the contents of the archive in a directory and then zipping that dir as a whole.

Results can be seen here: https://github.com/almarklein/wgpu-native/actions/runs/10703101259 (can others access these?)

One example:

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,17 +56,17 @@ package: lib-native lib-native-release
 		mkdir dist/$$ARCHIVEDIR; \
 		mkdir dist/$$ARCHIVEDIR/include; \
 		mkdir dist/$$ARCHIVEDIR/lib; \
-		cp ./dist/wgpu-native-git-tag  			dist/$$ARCHIVEDIR | true; \
-		cp ./ffi/webgpu-headers/webgpu.h  		dist/$$ARCHIVEDIR/include | true; \
-		cp ./ffi/wgpu.h  						dist/$$ARCHIVEDIR/include | true; \
-		cp ./$$LIBDIR/libwgpu_native.os  		dist/$$ARCHIVEDIR/lib | true; \
-		cp ./$$LIBDIR/libwgpu_native.dylib  	dist/$$ARCHIVEDIR/lib | true; \
-		cp ./$$LIBDIR/libwgpu_native.a  		dist/$$ARCHIVEDIR/lib | true; \
-		cp ./$$LIBDIR/libwgpu_native.dll.a  	dist/$$ARCHIVEDIR/lib | true; \
-		cp ./$$LIBDIR/wgpu_native.dll  			dist/$$ARCHIVEDIR/lib | true; \
-		cp ./$$LIBDIR/wgpu_native.lib  			dist/$$ARCHIVEDIR/lib | true; \
-		cp ./$$LIBDIR/wgpu_native.dll.lib  		dist/$$ARCHIVEDIR/lib | true; \
-		cp ./$$LIBDIR/wgpu_native.pdb  			dist/$$ARCHIVEDIR/lib | true; \
+		cp ./dist/wgpu-native-git-tag         dist/$$ARCHIVEDIR | true; \
+		cp ./ffi/webgpu-headers/webgpu.h      dist/$$ARCHIVEDIR/include | true; \
+		cp ./ffi/wgpu.h                       dist/$$ARCHIVEDIR/include | true; \
+		cp ./$$LIBDIR/libwgpu_native.os       dist/$$ARCHIVEDIR/lib | true; \
+		cp ./$$LIBDIR/libwgpu_native.dylib    dist/$$ARCHIVEDIR/lib | true; \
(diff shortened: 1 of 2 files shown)


almarklein:
cc @Korijn

I don't have a lot of experience with makefiles and shell scripts, so I would not be surprised if there are better ways to do things. E.g. I hate the `cd ../..`, but `popd`  was not available on one of the builds (IIRC the android build).

rajveermalviya (approved):
LGTM! Feel free to merge.

Though just noticed that ios archives (and the action name) have prefix `iOS` instead of `ios` making it inconsistent with other builds, would be good to have that fixed.

almarklein:
> Though just noticed that ios archives (and the action name) have prefix iOS instead of ios making it inconsistent with other builds, would be good to have that fixed.

Will do that in a new pr

almarklein:
Oh, I misunderstood your comment. Naming the archive to be lowercase is something that makes sense doing in this pr :)

almarklein:
New builds on my fork: https://github.com/almarklein/wgpu-native/actions/runs/10717624628

Makefile
```diff
@@ -45,20 +45,36 @@ endif
...
+		cp ./ffi/webgpu-headers/webgpu.h      dist/$$ARCHIVEDIR/include | true; \
+		cp ./ffi/wgpu.h                       dist/$$ARCHIVEDIR/include | true; \
+		cp ./$$LIBDIR/libwgpu_native.os       dist/$$ARCHIVEDIR/lib | true; \
+		cp ./$$LIBDIR/libwgpu_native.dylib    dist/$$ARCHIVEDIR/lib | true; \
+		cp ./$$LIBDIR/libwgpu_native.a        dist/$$ARCHIVEDIR/lib | true; \
+		cp ./$$LIBDIR/libwgpu_native.dll.a    dist/$$ARCHIVEDIR/lib | true; \
+		cp ./$$LIBDIR/wgpu_native.dll         dist/$$ARCHIVEDIR/lib | true; \
+		cp ./$$LIBDIR/wgpu_native.lib         dist/$$ARCHIVEDIR/lib | true; \
+		cp ./$$LIBDIR/wgpu_native.dll.lib     dist/$$ARCHIVEDIR/lib | true; \
+		cp ./$$LIBDIR/wgpu_native.pdb         dist/$$ARCHIVEDIR/lib | true; \
```
Korijn: I just wonder why the `| true`? We want to know when this fails right?

almarklein: Because many of these files are platform-specific. Can also do some if-triage instead. Yeah, let me change that to triage over different platforms and make it explicit.

almarklein: done

almarklein:
https://github.com/almarklein/wgpu-native/actions/runs/10722743457

Sahnvour:
Hi, i'm sorry if I didn't make it clear in the issue, but an intermediate `webgpu/` folder is missing in this case to be entirely useful.

In meta-build systems (cmake, meson, conan...) we'd want to do something like
`add_include_path('path/to/wgpu-unzipped/include')`
where `include/` still contains a `webgpu/` subfolder, so that other libraries depending on it can do `#include <webgpu/webgpu.h>`.

And if for any reason the project consuming wegpu wanted to use `#include <webgpu.h>` instead, they could do `add_include_path('path/to/wgpu-unzipped/include/webgpu')`: this way either option can be made to work.

almarklein:
> [...] where `include/` still contains a `webgpu/` subfolder, so that other libraries depending on it can do `#include <webgpu/webgpu.h>`.

For my understanding, why the need for the subfolder? The name "webgpu.h" seems unique enough by itself. I.e. can't these simply use `#include <webgpu.h>`?

Sahnvour:
> For my understanding, why the need for the subfolder? The name "webgpu.h" seems unique enough by itself. I.e. can't these simply use `#include <webgpu.h>`?

They can, but apparently the consensus is to use `<webgpu/webgpu.h>` in widely used libraries (that are often consumed via package managers and can't be manually edited by clients), see for example:
https://github.com/ocornut/imgui/blob/67cd4ead6505b3386d4ef9c713c98546e86e26ca/backends/imgui_impl_wgpu.h#L21
https://github.com/floooh/sokol/blob/1eb96dd0f96b9ea73065f9078244c2255c2b75d9/sokol_gfx.h#L4480

PR #418, opened 2024-09-08
Fix: Get correct adapter info consistent with `wgpuAdapterGetProperties`

lalitshankarch:
`wgpuAdapterGetInfo` was returning incorrect adapter info, which was inconsistent with the previous version (`wgpuAdapterGetProperties`).

With `wgpuAdapterGetProperties`:

```
vendorID: 4098
vendorName: AMD proprietary driver
architecture: 
deviceID: 5686
name: AMD Radeon(TM) Graphics
driverDescription: 24.3.1 (AMD proprietary shader compiler)
adapterType: 0x1
backendType: 0x6
```

With `wgpuAdapterGetInfo`:

```
vendorID: 4098
vendorName: 0x1002
architecture:
deviceID: 5686
name: 0x1636
driverDescription: AMD Radeon(TM) Graphics
adapterType: 0x1
backendType: 0x6
```

This is what `AdapterInfo` struct contains:

```rust
AdapterInfo {
    name: "AMD Radeon(TM) Graphics",
    vendor: 4098,
    device_type: IntegratedGpu,
    driver: "AMD proprietary driver",
    driver_info: "24.3.1 (AMD proprietary shader compiler)",
    backend: Vulkan,
}
```

And this is the output after my patch:

```
vendorID: 4098
vendorName: AMD proprietary driver
architecture:
deviceID: 5686
name: AMD Radeon(TM) Graphics
driverDescription: 24.3.1 (AMD proprietary shader compiler)
adapterType: 0x1
backendType: 0x6
```

All outputs were fetched from C++.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -699,14 +699,14 @@ pub unsafe extern "C" fn wgpuAdapterGetInfo(
         Err(err) => handle_error_fatal(err, "wgpuAdapterGetInfo"),
     };
 
-    info.vendor = CString::new(format!("{:#x}", result.vendor))
+    info.vendor = CString::new(result.driver)
         .unwrap()
         .into_raw();
     info.architecture = CString::default().into_raw();
-    info.device = CString::new(format!("{:#x}", result.device))
+    info.device = CString::new(result.name)
         .unwrap()
         .into_raw();
-    info.description = CString::new(result.name).unwrap().into_raw();
+    info.description = CString::new(result.driver_info).unwrap().into_raw();
     info.backendType = map_backend_type(result.backend);
     info.adapterType = map_adapter_type(result.device_type);
     info.vendorID = result.vendor;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -699,13 +699,9 @@ pub unsafe extern "C" fn wgpuAdapterGetInfo(
         Err(err) => handle_error_fatal(err, "wgpuAdapterGetInfo"),
     };
 
-    info.vendor = CString::new(result.driver)
-        .unwrap()
-        .into_raw();
+    info.vendor = CString::new(result.driver).unwrap().into_raw();
     info.architecture = CString::default().into_raw();
-    info.device = CString::new(result.name)
-        .unwrap()
-        .into_raw();
+    info.device = CString::new(result.name).unwrap().into_raw();
     info.description = CString::new(result.driver_info).unwrap().into_raw();
     info.backendType = map_backend_type(result.backend);
     info.adapterType = map_adapter_type(result.device_type);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -699,14 +699,10 @@ pub unsafe extern "C" fn wgpuAdapterGetInfo(
         Err(err) => handle_error_fatal(err, "wgpuAdapterGetInfo"),
     };
 
-    info.vendor = CString::new(format!("{:#x}", result.vendor))
-        .unwrap()
-        .into_raw();
+    info.vendor = CString::new(result.driver).unwrap().into_raw();
     info.architecture = CString::default().into_raw();
```
rajveermalviya:
```suggestion
    info.architecture = CString::default().into_raw(); // TODO(webgpu.h)
```
Let also ressurect the `TODO` comment for the `architecture`, from https://github.com/gfx-rs/wgpu-native/blob/a1ede115d0d669b88b34447e58752335cb882e61/src/lib.rs#L715

rajveermalviya (approved):
Thanks for catching it and pursuing to fix it!
LGTM!

almarklein:
I pushed a new tag/release.

almarklein:
Oh wait, I'm not sure if these changes are correct. This is what the WebGPU docs say:

image


So, `description`  should map to name, like it did, and `device`  should be the `device_id`.

rajveermalviya:
So, there are explicit fields for [`deviceId` and `vendorId`](https://webgpu-native.github.io/webgpu-headers/structWGPUAdapterInfo.html#a724155b0397fe85757eadea3ffa3df03) in `webgpu.h`, and the JS spec for `device` string says that it should be a vendor-specific adapter so this does seem correct mapping. One that seems off is the `vendor` string which we currently map to [`AdapterInfo.driver`](https://wgpu.rs/doc/wgpu/struct.AdapterInfo.html#structfield.driver), this could maybe changed to a todo and return an empty string.

I attempted to check what Dawn maps these strings to and discovered that wgpu-native and Dawn are results are similar:

| wgpu-native | Dawn |
| - | - |
| Screenshot 2024-09-09 at 17 29 00 | Screenshot 2024-09-09 at 17 17 39 |

| wgpu-native | Dawn |
| - | - |
| Screenshot 2024-09-09 at 17 31 15 | Screenshot 2024-09-09 at 17 30 45 |

almarklein:
Aha, thanks for the explanation. I am/was a in part confused by the WebGPU docs saying "'device' is a string that may be a PCI device ID.". That "may" is important, because in our case its not 😉 

*edit: In wgpu-py we need to map this to a WebGPU compliant dict, and this part feels like its not yet well-defined yet*

almarklein:
For future reference, in Chrome I currently get a rather empty dict:

image

rajveermalviya:
Yeah I saw that too, it's probably to prevent fingerprinting.

lalitshankarch:
The PCI ID is given by `DeviceID` BTW 😄.

While creating this PR I had to do a double take and match up with Dawn to check if I was correct or not. 😂

As for the `AdapterInfo.driver`, it's supposed to be the driver vendor.

PR #419, opened 2024-09-11
In binary artifacts, nest includes deeper

almarklein:
Follow-up on #417

@Sahnvour, should the lib-dir also be nested?

PR #420, opened 2024-09-12
actions: avoid relying on included ndk, setup manually

rajveermalviya:
Fixes Android workflow failures — https://github.com/gfx-rs/wgpu-native/pull/419#issuecomment-2345456934

While the exact root cause of the failures related to the Android NDK version included in the GitHub Actions images is unclear, we've encountered similar issues before. Therefore, it's more reliable to avoid depending on the NDK versions provided by GitHub.

This update now uses the latest LTS NDK version (`27.1.12297006`) — https://developer.android.com/ndk/downloads#lts-downloads

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -20,6 +20,7 @@ name: CD
 
 env:
   CACHE_SUFFIX: a
+  WGPU_ANDROID_NDK_VERSION: r27b # 27.1.12297006
 
 on:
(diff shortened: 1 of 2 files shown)


rajveermalviya:
Resulting CD builds — https://github.com/rajveermalviya/wgpu-native/actions/runs/10827876454

almarklein (approved):
Thanks for the quick fix!

PR #424, opened 2024-09-15
Change include path in `wgpu.h`

eliemichel:
The generated release zips files are organized as follows:

```
 - include/
   - webgpu/
     - webgpu.h
   - wgpu/
     - wgpu.h
 - lib/
   - ...
```

However, `wgpu.h` includes `webgpu.h` with `#include "webgpu.h"`, which forces to add `include/webgpu/` to the include directories, while it feels more idiomatic to only include `include/`. Other implementations also have `webgpu.h` in a `include/webgpu/` subfolder so I suggest leaving this so, leading to two possible fixes:

 - **Option A:** Change the include path in `wgpu.h` (this PR)
 - **Option B:** Distribute `wgpu.h` in `include/webgpu/` instead of `include/wgpu/` (what I had been doing manually in my distributions so far) -> edit: alternative PR for this option: https://github.com/gfx-rs/wgpu-native/pull/425

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -1,7 +1,7 @@
 #ifndef WGPU_H_
 #define WGPU_H_
 
-#include "webgpu.h"
+#include <webgpu/webgpu.h>
 
 typedef enum WGPUNativeSType {
     // Start at 0003 since that's allocated range for wgpu-native


almarklein:
@rajveermalviya and @Sahnvour thoughts?

I don't really care personally (I made wgpu-py find the files regardless of folder structure 😅).

rajveermalviya:
Thanks for catching this issue, @eliemichel!

Updating the checked-in file directly would cause problems with bindgen, and since this only impacts package builds, the best approach (for now) is to update the Makefile instead:

```diff
diff --git a/Makefile b/Makefile
index 6a5546e..ddea760 100644
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,7 @@ package: lib-native lib-native-release
                cp ./dist/wgpu-native-git-tag                 dist/$$ARCHIVEDIR; \
                cp ./ffi/webgpu-headers/webgpu.h              dist/$$ARCHIVEDIR/include/webgpu; \
                cp ./ffi/wgpu.h                               dist/$$ARCHIVEDIR/include/wgpu; \
+               sed -i '' 's/"webgpu.h"/<webgpu\/webgpu.h>/'  dist/$$ARCHIVEDIR/include/wgpu/wgpu.h; \
                if [ $(OS_NAME) = linux ]; then \
                        cp ./$$LIBDIR/libwgpu_native.so           dist/$$ARCHIVEDIR/lib; \
                        cp ./$$LIBDIR/libwgpu_native.a            dist/$$ARCHIVEDIR/lib; \
```

For reference, the folder structure was updated in #417 and #419, but we missed updating the include path because there’s no testing infrastructure for the built archives at the moment.

almarklein:
I know I said I don't care ... but changing the contents of a file in the makefile is rather unexpected and may cause confusion.

What about option B? Keeping `webgpu.h` and `wgpu.h` together does feel nice and simple, and `wgpu.h` basically defines 'extensions' so webgpu, so it kinda makes sense (to me)?

Sahnvour:
IMHO option B is simpler and more intuitive, I don't think it causes any trouble to have both headers in the same folder.

eliemichel:
If Option B feels like a better solution, here is an alternative PR (which indeed only affects the Makefile): https://github.com/gfx-rs/wgpu-native/pull/425

almarklein:
Thanks @eliemichel! Closing in favor of #425

PR #425, opened 2024-09-17
Distribute wgpu.h in include/webgpu

eliemichel:
This is the "**Option B**" mentionned as a potential alternative to #424


almarklein (approved):

almarklein:
@rajveermalviya are you ok with "option B"?

rajveermalviya:
Yes, now that I remember, the webgpu-native generator also expects similar structure for impl specific headers:
https://github.com/webgpu-native/webgpu-headers/blob/2b5974750be7b2545d842ecec02e585022081952/gen/cheader.tmpl#L61-L66

PR #427, opened 2024-09-19
Update library to latest webgpu-native headers

PJB3005:
This goes up to https://github.com/webgpu-native/webgpu-headers/commit/6f549cc3409663303b06f50483b9ebf9d1cf19dd
Things I *didn't* do:

* I didn't update the library to make sure "instance dropped" callback error codes are guaranteed to happen, like they seem to be in Dawn.

List of changes (roughly in order of header commits):

* Various enum and struct renames
* Updated callbacks to use the new `*CallbackInfo` structs and 2-userdata system. Also updated functions to return `WGPUFuture`, though the `WGPUFuture` thing is just stubbed out at the moment as I don't think wgpu-core has the necessary functionality for it. `wgpuInstanceWaitAny` is `unimplemented!()`
* `DepthClipControl` merged into `PrimitiveState`, related code simplified.
* Updated `depthWriteEnabled` to use an optional bool, mostly matters due to added validation.
* Add TODOs for missing features (sliced 3D compressed textures)
* `*Reference()` became `*AddRef()`
* Added `unorm10-10-10-2` vertex format
* Usage field in `TextureViewDescriptor`, just used for validation as wgpu-core doesn't allow specifying it anyways.
* Removed `maxInterStageShaderComponents`
* Added `clang_macro_fallback()` to bindgen config, since the headers switched to using `UINT32_MAX` etc. `UINT64_MAX` still doesn't work so I had to manually define those.
* Renamed flags enums. Added a conversion helper function to convert them from u64 -> u32 for mapping. (means added direct dependency on bitflags crate)
* Removed device argument from (unimplemented) `wgpuGetProcAddress`
* Suboptimal surface texture acquisition moved to enum return value, was easy since wgpu-core already returns it like that.
* "Undefined" present mode added, it just selects FIFO.
* `*EnumerateFeatures()` has been replaced with `*GetFeatures()`.
* All strings are now passed as `WGPUStringView`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -157,6 +157,7 @@ log = "0.4"
 thiserror = "1"
 parking_lot = "0.12"
 smallvec = "1"
+bitflags = "2.6.0"
 
 [build-dependencies]
 bindgen = "0.70"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -157,7 +157,7 @@ log = "0.4"
 thiserror = "1"
 parking_lot = "0.12"
 smallvec = "1"
-bitflags = "2.6.0"
+bitflags = "2"
 
 [build-dependencies]
 bindgen = "0.70"
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -293,7 +293,7 @@ pub fn map_instance_descriptor(
             backends: map_instance_backend_flags(extras.backends as native::WGPUInstanceBackend),
             dx12_shader_compiler,
             gles_minor_version: map_gles3_minor_version(extras.gles3MinorVersion),
-            flags: match extras.flags as native::WGPUInstanceFlag {
+            flags: match extras.flags {
                 native::WGPUInstanceFlag_Default => wgt::InstanceFlags::default(),
                 flags => map_instance_flags(flags),
             },
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -1,5 +1,5 @@
 use crate::utils::{make_slice, ptr_into_label, ptr_into_pathbuf};
-use crate::{follow_chain, map_enum};
+use crate::{follow_chain, map_enum, new_userdata};
 use crate::{native, UncapturedErrorCallback};
 use std::num::{NonZeroIsize, NonZeroU32, NonZeroU64};
 use std::ptr::NonNull;
@@ -164,7 +164,8 @@ map_enum!(
     Sint32,
     Sint32x2,
     Sint32x3,
-    Sint32x4
+    Sint32x4,
+    Unorm10_10_10_2
 );
 
 #[cfg(feature = "glsl")]
@@ -208,9 +209,10 @@ map_enum!(
     ReadWrite
 );
 
-pub const WGPU_WHOLE_SIZE: ::std::os::raw::c_ulonglong = native::WGPU_WHOLE_SIZE as _;
-pub const WGPU_LIMIT_U64_UNDEFINED: ::std::os::raw::c_ulonglong =
-    native::WGPU_LIMIT_U64_UNDEFINED as _;
+// These are defined as UINT64_MAX in the header, but bindgen currently can't process that define.
+// See https://github.com/rust-lang/rust-bindgen/issues/2822
+pub const WGPU_WHOLE_SIZE: u64 = u64::MAX;
+pub const WGPU_LIMIT_U64_UNDEFINED: u64 = u64::MAX;
 // it's SIZE_MAX in headers but it's not available in some compilers
 pub const WGPU_WHOLE_MAP_SIZE: usize = usize::MAX;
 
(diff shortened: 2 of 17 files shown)


PJB3005:
Remaining Windows CI failure is caused by https://github.com/webgpu-native/webgpu-headers/issues/339

almarklein (commented):
LGTM but maybe @rajveermalviya can have a look at the Rust code.

src/conv.rs
```diff
@@ -1660,13 +1648,18 @@ pub fn map_surface_configuration(
     config: &native::WGPUSurfaceConfiguration,
     extras: Option<&native::WGPUSurfaceConfigurationExtras>,
 ) -> wgt::SurfaceConfiguration<Vec<wgt::TextureFormat>> {
+    let present_mode = match config.presentMode {
+        native::WGPUPresentMode_Undefined => wgt::PresentMode::Fifo,
+        _ => map_present_mode(config.presentMode),
+    };
```
eliemichel: I believe this `match` should be part of `map_present_mode` rather than being added on top of it

src/conv.rs
```diff
@@ -1701,3 +1694,11 @@ pub fn map_adapter_type(device_type: wgt::DeviceType) -> native::WGPUAdapterType
         wgt::DeviceType::Cpu => native::WGPUAdapterType_CPU,
     }
 }
+
+pub fn from_u64_bits<T: bitflags::Flags<Bits = u32>>(value: u64) -> Option<T> {
```
eliemichel, line 1711: (Note: Ideally bitfields would switch to `u64` in upstream `wgpu` instead of needing a conversion here)

Cargo.toml
```diff
@@ -157,6 +157,7 @@ log = "0.4"
 thiserror = "1"
 parking_lot = "0.12"
 smallvec = "1"
+bitflags = "2.6.0"
```
eliemichel: why is this needed?

eliemichel (approved):
Nice job! I started looking at how to implement futures in `wgpu` a couple of days ago, nothing obvious but I'll share if I figure sth out!

Cargo.toml
PJB3005: It's to be able to refer to `bitflags` types for the `from_u64_bits` helper.

src/conv.rs
PJB3005: I did this here because I wasn't sure if I could cleanly add that to `map_present_mode`, as it's defined with a `map_enum!` macro.

eliemichel (approved):

eliemichel:
For WGPUFuture, we need to have access to submission index (and return that in the future's id). Here is the beginning of an attempt at adding this upstream: https://github.com/gfx-rs/wgpu/pull/6360

src/utils.rs
```diff
@@ -268,6 +270,69 @@ macro_rules! map_enum {
...
+            0 => Some(""),
+            _ => panic!("Null address to WGPUStringView!"),
+        }
+    } else {
+        let bytes = match string_view.length {
+            crate::conv::WGPU_STRLEN => CStr::from_ptr(string_view.data).to_bytes(),
+            _ => make_slice(string_view.data as *const u8, string_view.length),
+        };
+
+        Some(std::str::from_utf8_unchecked(bytes))
```
PJB3005, line 327: Important to note: this doesn't check whether strings passed into the library are valid UTF-8. I previously many locations did, but I figured this wasn't necessary.

eliemichel:
FYI I started trying to build upon this PR and upgrade wgpu to my [PR that adds submission indices for futures](https://github.com/gfx-rs/wgpu/pull/6360), WIP is here: https://github.com/eliemichel/wgpu-native/tree/eliemichel/2024-10-10-upgrade-wgpu-futures

Ongoing work: I need to figure out what replaced the `DynComputePass` and `DynRenderPass` types in newest wgpu.

ygdrasil-io:
There is a missing fonction `wgpuGetInstanceFeatures` on the rust code but describe on yaml as :
```yaml
  - name: get_instance_features
    doc: Query the supported instance features
    args:
      - name: features
        doc: The supported instance features
        type: struct.instance_features
        pointer: mutable
```

eliemichel:
> FYI I started trying to build upon this PR and upgrade wgpu

This is now ready, PR: #441

PJB3005:
@ygdrasil-io done, thanks for pointing that out. Though note it's not very useful since WaitAny isn't implemented yet.

src/conv.rs
PJB3005: Ended up doing this properly due to more incoming changes in the latest headers.

eliemichel:
@rajveermalviya Any specific reason why this PR is on hold? (You not having had time to review it yet is a perfectly fine reason, just trying to figure out if there is a technical blocker I could help with!)

eliemichel:
Merged latest commits into #441 and did some extra upgrade of webgpu.h, leading to #445, still hoping for this to get merged!

PJB3005:
Closing since the other PR was merged.

PR #430, opened 2024-09-24
Implement wgpuInstanceProcessEvents

eliemichel:
Polling all devices

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2662,7 +2662,7 @@ pub unsafe extern "C" fn wgpuInstanceCreateSurface(
 pub unsafe extern "C" fn wgpuInstanceProcessEvents(instance: native::WGPUInstance) {
     let instance = instance.as_ref().expect("invalid instance");
     let context = &instance.context;
-    
+
     match context.poll_all_devices(false) {
         Ok(_queue_empty) => (),
         Err(cause) => {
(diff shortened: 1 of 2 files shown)


almarklein (approved):
Thanks!

almarklein:
I happen to look into enabling async support for wgpu-py now, so the timing couldn't be better 😄 

@rajveermalviya how does `wgpuInstanceProcessEvents` (which is in `webgpu.h`) relate to `wgpuDevicePoll` (which is in `wgpu.h`, [here](https://github.com/gfx-rs/wgpu-native/blob/58c83560015701e28b8850ddd4515fcf0a4dd60d/src/lib.rs#L4197))? I presume the latter will be deprecated at some point?

This PR makes`wgpuInstanceProcessEvents` polls all devices, but is that enough? Does the current call do all the things mentioned here: https://github.com/webgpu-native/webgpu-headers/issues/117?

And then there's https://github.com/webgpu-native/webgpu-headers/issues/199 and  https://github.com/webgpu-native/webgpu-headers/pull/314 ... From what I understand, the Future API allows polling for callbacks precisely, while `wgpuInstanceProcessEvents` is a global "tick" (and does some extra things.

almarklein:
Some more info on my own questions:
* `wgpuInstanceProcessEvents` calls [context.poll_all_devices](https://docs.rs/wgpu-core/latest/wgpu_core/global/struct.Global.html#method.poll_all_devices)
* `wgpuDevicePoll` calls  [context.device_poll](https://docs.rs/wgpu-core/latest/wgpu_core/global/struct.Global.html#method.device_poll)

From what I understand the former is just more precise (allowing polling a specific device, and with a specific `submission_index`.

Issue #433, opened 2024-09-29
add webgpu.yml on release


ygdrasil-io:
Hello! I've started using the [webgpu.yml](https://github.com/webgpu-native/webgpu-headers/blob/main/webgpu.yml) to regenerate my Kotlin bindings, and it's definitely more convenient than working directly with the C headers. I was wondering if the community would be open to including this YAML file alongside the headers in each release ?

almarklein:
Makes sense to me! [This line](https://github.com/gfx-rs/wgpu-native/blob/58c83560015701e28b8850ddd4515fcf0a4dd60d/Makefile#L59) moves the `webgpu.h` into the archive, so adding a line there for the yml file should do the trick.

PR #435, opened 2024-10-06
include webgpu.yml on packaging

ygdrasil-io:
Related to https://github.com/gfx-rs/wgpu-native/issues/433. 
I created a mock version to test the packaging : https://github.com/ygdrasil-io/wgpu-native/releases/tag/v0.0.0.7

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,11 +53,11 @@ package: lib-native lib-native-release
 		LIBDIR=$(TARGET_DIR)/$$RELEASE; \
 		rm -r -f dist/$$ARCHIVEDIR; \
 		rm -f dist/$$ARCHIVEFILE; \
-		mkdir -p dist/$$ARCHIVEDIR/specifications; \
+		mkdir -p dist/$$ARCHIVEDIR/wgpu-native-meta; \
 		mkdir -p dist/$$ARCHIVEDIR/include/webgpu; \
 		mkdir -p dist/$$ARCHIVEDIR/lib; \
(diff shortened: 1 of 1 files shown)


eliemichel:
Nice addition, I only wonder whether "specifications" (plural) is the best wording.

almarklein:
Maybe just put it next to the `.h` file? Or in the root?

ygdrasil-io:
"There are only two hard things in Computer Science: cache invalidation and naming things."
-- Phil Karlton

I only added a directory to try to add semantics but the root directory is fine from my point of view.
Placing it in the "include" folder might be misleading, as this file won't be processed by a compiler.

So, root folder it is ?

almarklein:
> cache invalidation and naming things

Very true :)

> So, root folder it is ?

Actually, I gave this some thought and I prefer we create a subdir called `wgpu-native-meta`, and put both the `webgpu.yml` and `wgpu-native-git-tag` in there.

The reasoning:

The purpose of the subdirs is to make it possible to extract the archive in existing projects and have everything nicely in-place.

The one file we had in the root folder was no problem, because its name is unique enough. But now we add another file, and I plan to add files that contain a hash of the lib, and who knows what we'll add later. IMO it makes sense to cleanly put them in a folder that is easily identified.

For consumers that don't care about the structure, its easy enough to flatten it (we do something like this in wgpu-py).

almarklein (approved):

almarklein:
Thanks!

PR #437, opened 2024-10-10
implement `wgpuComputePassEncoderSetPushConstants`

zackgomez:
It was straightforward to implement.  I don't expect my changes to the example to be merged, I am looking for guidance on whether they should be reverted, moved to another example, or something else entirely.

Test Plan:
Modified the compute shader to use push constants.  Verified the output.

### DIFF
--- a/examples/push_constants/main.c
+++ b/examples/push_constants/main.c
@@ -1,0 +1,253 @@
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "framework.h"
+#include "webgpu-headers/webgpu.h"
+
+#define LOG_PREFIX "[push_constants]"
+
+static void handle_request_adapter(WGPURequestAdapterStatus status,
+                                   WGPUAdapter adapter, char const *message,
+                                   void *userdata) {
+  UNUSED(status)
+  UNUSED(message)
+  *(WGPUAdapter *)userdata = adapter;
+}
+static void handle_request_device(WGPURequestDeviceStatus status,
+                                  WGPUDevice device, char const *message,
+                                  void *userdata) {
+  UNUSED(status)
+  UNUSED(message)
+  *(WGPUDevice *)userdata = device;
+}
+static void handle_buffer_map(WGPUBufferMapAsyncStatus status, void *userdata) {
+  UNUSED(userdata)
+  printf(LOG_PREFIX " buffer_map status=%#.8x\n", status);
+}
+
+int main(int argc, char *argv[]) {
+  UNUSED(argc)
+  UNUSED(argv)
+  frmwrk_setup_logging(WGPULogLevel_Warn);
+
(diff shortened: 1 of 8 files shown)


rajveermalviya:
Let's keep the `compute` example as-is, but a separate example for `push_constants` would be great to have. Also seems like `cargo fmt` CI is not happy.

almarklein:
cc @fyellin since you implemented render push constants in wgpu-py.

zackgomez:
rebased and added examples/push_constants

Also noting that I used my fork of wgpu with these PRs in a project and they all worked as expected

examples/push_constants/main.c
```diff
@@ -0,0 +1,245 @@
+#include "framework.h"
+#include "webgpu-headers/webgpu.h"
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#define LOG_PREFIX "[compute]"
```
rajveermalviya:
```suggestion
#define LOG_PREFIX "[push_constants]"
```

examples/push_constants/main.c
```diff
@@ -0,0 +1,245 @@
...
+  wgpuCommandEncoderCopyBufferToBuffer(command_encoder, storage_buffer, 0,
+                                       staging_buffer, 0, numbers_size);
+  
+  WGPUCommandBuffer command_buffer = wgpuCommandEncoderFinish(
+      command_encoder, &(const WGPUCommandBufferDescriptor){
+                           .label = "command_buffer",
+                       });
+  assert(command_buffer);
+
+  wgpuQueueWriteBuffer(queue, storage_buffer, 0, &numbers, numbers_size);
```
rajveermalviya, line 226: This write is unnecessary, correct?

rajveermalviya (changes requested):
Looks great so far! Added couple of comments.

We could also consider porting over [this wgpu test](https://github.com/gfx-rs/wgpu/blob/trunk/tests/tests/push_constants.rs) so that we have a reference to track any future wgpu-core API changes and easily update the examples if needed.

rajveermalviya:
Apart from the example, the API changes LGTM (and tests correctly with current example code). If you need we could merge the API changes in this PR, and do the example in a separate PR.

examples/push_constants/main.c
zackgomez: applied

fyellin:
When will this code be available to use in wgpu-py?

almarklein:
> When will this code be available to use in wgpu-py?

@fyellin when a new version of wgpu-native is released, and wgpu-py updates to that release.

zackgomez:
pls merge

rajveermalviya (approved):
Thanks, Looks good! 

(I also added a commit to run the example in CI)

almarklein:
Thanks @zackgomez !

PR #450, opened 2025-01-10
meson: Install Cargo-generated lib directly

obiwac:
Currently, when building with Meson, an archive is created containing the archive generated by Cargo (so not directly the object files contained withing the Cargo-generated archive). This causes issues when linking and required an explicit `ranlib` or `ar -s` step to be issued before the library is usable:

```log
ld: archive member 'libwgpu_native.a' not a mach-o file in '.../libwgpu_native.a'
```

[It doesn't look like](https://mesonbuild.com/Reference-manual_functions.html#static_library) `static_library` really supports this kind of usecase.

This PR simply installs the static library artifact generated by Cargo directly instead. Unfortunately I don't think there's an elegant way to pass a meson library object to `pkg.generate` in this situation, so I've just set `libraries` manually (since it allows arbitrary strings). I'm by no means a Meson expert so do let me know if there's a better way that I'm overlooking :)

@KentaTheBugMaker pinging you as the original author of `meson.build` (#84) for your input.


almarklein:
This looks fine by me, but @KentaTheBugMaker understand this code way better.

PR #451, opened 2025-01-12
String passed to the callback must be null terminated

levydsa:
This is quite hard to spot. Stumbled upon it while trying to force `wgpuInstanceRequestAdapter` to return a message in the response.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2696,7 +2696,7 @@ pub unsafe extern "C" fn wgpuInstanceRequestAdapter(
                         callback(
                             native::WGPURequestAdapterStatus_Error,
                             std::ptr::null_mut(),
-                            "unsupported backend type: d3d11\0".as_ptr() as _,
+                            c"unsupported backend type: d3d11".as_ptr() as _,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2696,7 +2696,7 @@ pub unsafe extern "C" fn wgpuInstanceRequestAdapter(
                         callback(
                             native::WGPURequestAdapterStatus_Error,
                             std::ptr::null_mut(),
-                            "unsupported backend type: d3d11".as_ptr() as _,
+                            "unsupported backend type: d3d11\0".as_ptr() as _,
```
cwfitzgerald: Nit: this uses a c literal which has a \0 automatically added.

```suggestion
                            c"unsupported backend type: d3d11".as_ptr() as _,
```

cwfitzgerald (approved):
Looks good, have a nit

src/lib.rs
levydsa: I didn't knew this was a thing, awesome :)

PR #453, opened 2025-01-20
Use manylinux2014

almarklein:
This is about the CD build. It fails :)


hmaarrfk:
I've kinda already debugged this avenue:

1. `dnf` replaced `yum` in more recent centos's
2. the llvm included in centos7 is simply too old. 

I necro'ed https://github.com/pypa/manylinux/issues/1728#issuecomment-2602817305 asking for newer toolsets, but we'll see.


(I don't know why i spend time on pypi packages..... I already devote too much time helping out on coda-forge..... should spend my time more wisely)

hmaarrfk:
alternative in https://github.com/gfx-rs/wgpu-native/pull/454

almarklein:
Closing for now. The idea was to be able to support older Linux versions, but we're running into issues making this a bit of a rabit-hole.

Issue #452, opened 2025-01-19
Building for older versions of linux


hmaarrfk:
I was asked to open an issue to discuss my successes in building wgpu-native for older GLIBC standards.

I'm mostly here to report the success i've had at conda-forge building wgpu-native for libc 2.17 at [conda-forge](https://github.com/conda-forge/wgpu-native-feedstock/tree/main/recipe]

I was [told](https://github.com/pygfx/wgpu-py/issues/664#issuecomment-2599847979) that the reason that 2.28 is the minimum supported version is due to struggles with CIs and configurations that seemed to give trouble otherwise.

At conda-forge, the situation is a little special:

1. Very modern compilers are used (latest or second latest)
2. Very old (in this case 2.17) is targetted.

Some of this information is immortalized in our ci configuration files: https://github.com/conda-forge/wgpu-native-feedstock/blob/main/.ci_support/linux_64_.yaml You can see the gcc version and the stdlib version.

honestly, I struggle with opening these kinds of issues since I don't even know what platforms you want to target. If you are happy with the existing situation feel free to close without any comments.

I gave it a go: https://github.com/hmaarrfk/wgpu-native/pull/1 and noticed that it is the libclang that is just too old on the manylinux2014 image. At conda-forge, i notice through our azure logs that we are pulling in clang 19 so pretty modern.

I'm not sure if the rust community has a vendorered libclang that you can pull in your builds instead of that of the manylinux image you pull in https://github.com/gfx-rs/wgpu-native/blob/a3d1589014e4a1791e0c9928f79c24955d9fda62/.github/workflows/cd.yml#L226

cc: @Korijn 

Maybe this "existance proof" is good enough to give you ideas.

Korijn:
Providing some historical context:

* manylinux_2_28: current
* manylinux_2_24: https://github.com/gfx-rs/wgpu-native/pull/256
* manylinux2010: https://github.com/gfx-rs/wgpu-native/pull/27 (this is such a long time ago already wow)
* Before then it wasn't a standardized build environment (the binaries also weren't compliant with any standard)

I hope you can get some insight into why the image was changed by examining these threads.

hmaarrfk:
Those threads mostly hint that you guys are moving forward with the latest recommendations. 

That is fine

Here is a user that requests for manylinux2014 binary. 
https://github.com/gfx-rs/wgpu-native/pull/256#issuecomment-1546530583


[...]

PR #454, opened 2025-01-20
Build linux-x86-64 on a container that supports manylinux2014

hmaarrfk:
I don't know how to get cross compilation working for the arm container.

But this should get us going for slightly older manylinux targets

https://github.com/gfx-rs/wgpu-native/issues/452


hmaarrfk:
I'll let a maintainer "reopen" this PR if they are interested, but I am not interested in advocating for it since I believe this adds complexity and better dealt with at the PyPa level

PR #463, opened 2025-02-25
Fix header-implementation incongruencies

### LINKED ISSUE #462: wgpuAdapterGetInfo header-implementation incongruency
Hi all,

I recently updated my wgpu-native project to version v24.0.0.1. I noticed that the function `wgpuAdapterGetInfo` returns an invalid `wgpuStatus` enum value.

The rust function is defined as follows (both in the latest commit and in the source of the release v24.0.0.1):
```rust
pub unsafe extern "C" fn wgpuAdapterGetInfo(
    adapter: native::WGPUAdapter,
    info: Option<&mut native::WGPUAdapterInfo>,
)
```
Consider however the C `wgpuAdapterGetInfo` definition (both in the webgpu-headers repository, in the latest commit and in the header distributed in the release):
```c
WGPU_EXPORT WGPUStatus wgpuAdapterGetInfo(WGPUAdapter adapter, WGPUAdapterInfo * info) WGPU_FUNCTION_ATTRIBUTE;
```

The implementation of `wgpuAdapterGetInfo` does not return a status, leaving its return value undefined.

vicixdev:
Fixes #462.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -723,7 +723,7 @@ pub unsafe extern "C" fn wgpuAdapterGetLimits(
 pub unsafe extern "C" fn wgpuAdapterGetInfo(
     adapter: native::WGPUAdapter,
     info: Option<&mut native::WGPUAdapterInfo>,
-) {
+) -> native::WGPUStatus {
     let adapter = adapter.as_ref().expect("invalid adapter");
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
Thanks!

Issue #190, opened 2022-04-26
Add support for WGPU native features


MineGame159:
As the title says it would be good to add support for features / extensions WGPU adds on top of WebGPU for example push constants.

almarklein:
Update: we now have:
```
WGPUNativeFeature_PushConstants = 0x60000001,
WGPUNativeFeature_TextureAdapterSpecificFormatFeatures = 0x60000002,
WGPUNativeFeature_MultiDrawIndirect = 0x60000003,
WGPUNativeFeature_MultiDrawIndirectCount = 0x60000004,
WGPUNativeFeature_VertexWritableStorage = 0x60000005,
```

But that does not seem to be the complete set of native features as specified on https://docs.rs/wgpu/latest/wgpu/struct.Features.html

davnotdev:
I would like to work on this. Is there a specific convention or order for each native feature?

almarklein:
I'd say follow wgpu-core in case of doubt.

davnotdev:
I believe the current order doesn't match wgpu-core.

PR #477, opened 2025-04-27
upgrade to wgpu v25

ygdrasil-io:
Upgrade to wgpu v25 and update the Android NDK version to v28.
I built a version for testing :
https://github.com/ygdrasil-io/wgpu-native/releases/tag/v25.0.0.1

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -118,17 +118,17 @@ typedef enum WGPUNativeQueryType {
     WGPUNativeQueryType_Force32 = 0x7FFFFFFF
 } WGPUNativeQueryType WGPU_ENUM_ATTRIBUTE;
 
-typedef enum WGPUDirectXMaxShaderModel {
-    WGPUDirectXMaxShaderModel_V6_0 = 0x00000000,
-    WGPUDirectXMaxShaderModel_V6_1 = 0x00000001,
-    WGPUDirectXMaxShaderModel_V6_2 = 0x00000002,
-    WGPUDirectXMaxShaderModel_V6_3 = 0x00000003,
-    WGPUDirectXMaxShaderModel_V6_4 = 0x00000004,
-    WGPUDirectXMaxShaderModel_V6_5 = 0x00000005,
-    WGPUDirectXMaxShaderModel_V6_6 = 0x00000006,
-    WGPUDirectXMaxShaderModel_V6_7 = 0x00000007,
-    WGPUDirectXMaxShaderModel_Force32 = 0x7FFFFFFF
-} WGPUDirectXMaxShaderModel;
+typedef enum WGPUDxcMaxShaderModel {
+    WGPUDxcMaxShaderModel_V6_0 = 0x00000000,
+    WGPUDxcMaxShaderModel_V6_1 = 0x00000001,
+    WGPUDxcMaxShaderModel_V6_2 = 0x00000002,
+    WGPUDxcMaxShaderModel_V6_3 = 0x00000003,
+    WGPUDxcMaxShaderModel_V6_4 = 0x00000004,
+    WGPUDxcMaxShaderModel_V6_5 = 0x00000005,
+    WGPUDxcMaxShaderModel_V6_6 = 0x00000006,
+    WGPUDxcMaxShaderModel_V6_7 = 0x00000007,
+    WGPUDxcMaxShaderModel_Force32 = 0x7FFFFFFF
+} WGPUDxcMaxShaderModel;
 
 typedef enum WGPUGLFenceBehaviour {
     WGPUGLFenceBehaviour_Normal = 0x00000000,
@@ -145,7 +145,7 @@ typedef struct WGPUInstanceExtras {
     WGPUGLFenceBehaviour glFenceBehaviour;
     WGPUStringView dxilPath;
     WGPUStringView dxcPath;
-    WGPUDirectXMaxShaderModel dxcMaxShaderModel;
+    WGPUDxcMaxShaderModel dxcMaxShaderModel;
 } WGPUInstanceExtras;
 
 typedef struct WGPUDeviceExtras {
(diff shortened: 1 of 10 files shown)


src/conv.rs
```diff
@@ -296,6 +296,7 @@ pub unsafe fn map_instance_descriptor(
                 (Some(dxil_path), Some(dxc_path)) => wgt::Dx12Compiler::DynamicDxc {
                     dxil_path: dxil_path.to_string(),
                     dxc_path: dxc_path.to_string(),
+                    max_shader_model: wgt::DxcShaderModel::V6_7,
```
cwfitzgerald: We should add this as a member in `extras`

src/conv.rs
```diff
@@ -307,10 +308,12 @@ pub unsafe fn map_instance_descriptor(
             backend_options: wgt::BackendOptions {
                 gl: wgt::GlBackendOptions {
                     gles_minor_version: map_gles3_minor_version(extras.gles3MinorVersion),
+                    fence_behavior: Default::default(),
```
cwfitzgerald: This as well

src/conv.rs
```diff
@@ -1142,11 +1144,12 @@ pub fn features_to_native(features: wgt::Features) -> Vec<native::WGPUFeatureNam
...
-        .contains(wgt::Features::UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING)
-    {
-        temp.push(native::WGPUNativeFeature_UniformBufferAndStorageTextureArrayNonUniformIndexing);
-    }
+    // TODO: fix this, UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING is not supported anymore https://github.com/gfx-rs/wgpu/issues/4407
+    // if features
+    //     .contains(wgt::Features::UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING)
+    // {
+    //     temp.push(native::WGPUNativeFeature_UniformBufferAndStorageTextureArrayNonUniformIndexing);
+    // }
```
cwfitzgerald, line 1175: Could you file an issue about fixing these, or include the fix in the PR?

src/lib.rs
```diff
@@ -4282,14 +4279,15 @@ pub unsafe extern "C" fn wgpuDevicePoll(
...
+            None => wgt::PollType::Wait,
         },
-        false => wgt::Maintain::Poll,
+        false => wgt::PollType::Poll,
     };
 
     match context.device_poll(device_id, maintain) {
-        Ok(queue_empty) => queue_empty,
+        // TODO check if we need to check the PollStatus to return true or false
+        Ok(_) => true,
```
cwfitzgerald: yes, this needs to match if PollStatus == QueueEmpty

cwfitzgerald (changes requested):
Looking good, some concerns

cwfitzgerald:
With the new behavior in features, I think we can use a single unconditional dep on wgc with the dx12, gles, vulkan, and metal features.

src/conv.rs
almarklein: We can simply remove it from `wgpu.h`.

ygdrasil-io: I think we should review all the enumerations at once, as some are commented. Is there already an issue for tacking this?

src/lib.rs
ygdrasil-io: Should we log something when it does not match ?

src/conv.rs
almarklein: Yes, would be good to review the feature enums 👌   Issue #190 seems related.

src/lib.rs
cwfitzgerald: I mean, `matches!(poll_status, QueueEmpty)` instead of always true

ygdrasil-io: I get it, but should we add a log to explain what is happening when we return false ?

ygdrasil-io: fixed

src/conv.rs
ygdrasil-io: If that's alright for everybody, I'm keeping it as is, as this will be fixed in #190.

src/conv.rs
ygdrasil-io: fixed

src/conv.rs
ygdrasil-io: fixed

ygdrasil-io:
@cwfitzgerald are you expecting more changes ?

cwfitzgerald:
Just didn't realize this was updated. I'll see if I can take a look at it soon-ish

ffi/wgpu.h
```diff
@@ -118,14 +118,34 @@ typedef enum WGPUNativeQueryType {
     WGPUNativeQueryType_Force32 = 0x7FFFFFFF
 } WGPUNativeQueryType WGPU_ENUM_ATTRIBUTE;
 
+typedef enum WGPUDirectXMaxShaderModel {
```
cwfitzgerald: I'm not sure we should deviate from `wgpu`'s name here, without good reason - this is specifically about the max shader model of the DXC exe the user is handing to wgpu. I think this should either be `WGPUDx12ShaderModel` or `WGPUDxcMaxShaderModel`

cwfitzgerald (approved):
Good to go after one comment and a rebase.

ffi/wgpu.h
ygdrasil-io: fixed

ygdrasil-io:
- Rebase done
- WGPUDirectXMaxShaderModel changed to WGPUDxcMaxShaderModel
- Upgraded to latest minor version

@cwfitzgerald or @almarklein If that's all right, can you merge and do a release please?

almarklein (approved):

almarklein:
Thanks! 🚀

PR #455, opened 2025-01-25
upgrade to wgpu 24.0.0

ygdrasil-io:
This PR is based on https://github.com/gfx-rs/wgpu-native/pull/445 to upgrade wgpu to 24.0.0 and remove the memory leak.

Feel free to comment regarding the last changes.

I found one memory leak here: https://github.com/eliemichel/wgpu-native/commit/452b21e84cd7cbd65f7ddf55726dfaaab99e5ff0, but there is another one.

After 5-10 minutes of running, this decreases the memory leak by 50%, but it is still leaking, while the same sample is memory stable on the wgpu Rust repository.

PR #478, opened 2025-05-02
ffi/wgpu.h: Typo in param name & GLSL shader source struct

obiwac:
Presumably this got missed in https://github.com/gfx-rs/wgpu-native/pull/455

I'm going through updating my bindings to v0.24.0, so I might come across other such typos which I'll add to this PR. So feel free to let this sit for a few days :)

I am unsure of if there's a difference between `WGPUShaderModuleDescriptorSpirV` and `WGPUShaderSourceSPIRV`, so I left this as is. `wgpuDeviceCreateShaderModuleSpirV` is to be replaced by `create_shader_module_passthrough` in v25 anyway.


almarklein (approved):
Nice catch!

Issue #493, opened 2025-06-26
MacOS lib-native build fails with unknown types


brandonpelfrey:
Currently on the latest release, I am experiencing build failures given the provided build instructions. I did not see an existing issue open for this. I suspect bindgen is not including the correct headers. These types (uint32_t etc) are all definitely in the stdint.h header. Please let me know if I can provide more information.

```
wgpu-native % git reset --hard v25.0.2.1

wgpu-native % rustup update

wgpu-native % cargo update

wgpu-native % make lib-native
cargo build
   Compiling wgpu-native v0.0.0 (/Users/***/p/wgpu-native)
error: failed to run custom build command for `wgpu-native v0.0.0 (/Users/***/p/wgpu-native)`

Caused by:
  process didn't exit successfully: `/Users/***/p/wgpu-native/target/debug/build/wgpu-native-d506a8c2b2a9d0c6/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=ffi/webgpu-headers/webgpu.h
  cargo:rerun-if-changed=ffi/wgpu.h
  cargo:rerun-if-env-changed=TARGET
  cargo:rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS_aarch64-apple-darwin
  cargo:rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS_aarch64_apple_darwin
  cargo:rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS
  cargo:rerun-if-changed=ffi/wgpu.h

  --- stderr
  ffi/webgpu-headers/webgpu.h:96:9: error: unknown type name 'uint64_t'
  ffi/webgpu-headers/webgpu.h:97:9: error: unknown type name 'uint32_t'
  ffi/webgpu-headers/webgpu.h:1395:5: error: unknown type name 'uint32_t'
  ffi/webgpu-headers/webgpu.h:1396:5: error: unknown type name 'uint32_t'
  ffi/webgpu-headers/webgpu.h:1401:5: error: unknown type name 'uint32_t'
  ffi/webgpu-headers/webgpu.h:1403:5: error: unknown type name 'uint64_t'
  ffi/webgpu-headers/webgpu.h:1404:5: error: unknown type name 'uint64_t'
  ffi/webgpu-headers/webgpu.h:1419:5: error: unknown type name 'uint64_t'
[...]

PR #494, opened 2025-06-26
Explicitly pass MacOS SDK sysroot to bindgen arguments

brandonpelfrey:
Resolves issue in #493 . Bindgen is not including the proper MacOS SDK sysroot. Explicitly passing this resolves the build issue 

```sh
wgpu-native % ( make && make lib-native) 1>/dev/null && echo "Build success" || "Build failed"
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
    Finished `release` profile [optimized] target(s) in 0.03s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
Build success
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,11 +59,14 @@ jobs:
             os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
             # Macos
-          - name: Macos x86_64
+          - name: Macos 14 x86_64
             os: macos-14
             target: x86_64-apple-darwin
-          - name: Macos aarch64
+          - name: Macos 14 aarch64
             os: macos-14
+            target: aarch64-apple-darwin
+          - name: Macos 15 aarch64
+            os: macos-15
             target: aarch64-apple-darwin
             # Windows
           - name: Windows gnu x86_64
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -59,12 +59,15 @@ jobs:
...
             target: x86_64-unknown-linux-gnu
             # Macos
-          - name: Macos x86_64
+          - name: Macos 14 x86_64
             os: macos-14
             target: x86_64-apple-darwin
-          - name: Macos aarch64
+          - name: Macos 14 aarch64
             os: macos-14
             target: aarch64-apple-darwin
```
cwfitzgerald, line 67: Is there a pressing reason why we would keep these around?

cwfitzgerald (approved):
Looks good, one comment

.github/workflows/ci.yml
brandonpelfrey: I would probably still keep the MacOS 14 around for a while. EOL for MacOS 14 is late 2026.

brandonpelfrey:
Hi @cwfitzgerald apologies, but for some reason it looks like the older build matrix entries are being expected ("Macos aarch64", "Macos x86_64")? These don't exist after the new change now that we have explicitly named MacOS 14 and 15. This appears to be blocking the completion of the pipeline even though all the builds have completed. Do you know why it's still waiting on older steps to run?

cwfitzgerald:
This is because we mark all our CI jobs as required, so it's waiting for the job of that name to finish - I can just update the list, no worries about that.

PR #497, opened 2025-07-07
Fix read-only depth/stencil attachments

AlbinBernhardssonARM:
Read-only depth/stencil attachments must have load/store ops None (or _Undefined enum value in C API). However, when converting WGPURenderPassDepthStencilAttachment, WGPULoadOp_Undefined was being mapped to Some(wgc::command::LoadOp::Load) rather than None. Same for store op.


lisyarus:
Bumping this. The current code makes it impossible to use depth_read_only and stencil_read_only. This review fixes that.

almarklein (approved):

almarklein:
Thanks!

PR #504, opened 2025-08-09
Add binding number to related error messages

devprofile98:
I have added the binding number to the error message for easier debugging. So now the user can see which binding is failing.


almarklein (approved):

devprofile98:
@almarklein Hi, should I do anything for the workflow approval?

almarklein:
Thanks!

PR #509, opened 2025-08-26
Update to wgpu v27.0.0

melody-rs:
Updates wgpu-native to wgpu v27.0.0!
I also implemented some features commented out in `wgpu.h` but I plan to add more later after splitting up wgpu-native so `lib.rs` and `conv.rs` aren't so long

This PR was originally designed around v26, but v27 just released so I went ahead and updated it.
v27 support needs some work still- some of the new functionality in v27 needs to be exposed in `wgpu.h`. I'd like to do that in another PR after I've split up `conv.rs` and `lib.rs`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,23 +22,19 @@ resolver = "2"
 
 [workspace.dependencies.wgc]
 package = "wgpu-core"
-git = "https://github.com/gfx-rs/wgpu"
-tag = "v27.0.0"
+version = "27.0.1"
 
 [workspace.dependencies.wgt]
 package = "wgpu-types"
-git = "https://github.com/gfx-rs/wgpu"
-tag = "v27.0.0"
+version = "27.0.1"
 
 [workspace.dependencies.hal]
 package = "wgpu-hal"
-git = "https://github.com/gfx-rs/wgpu"
-tag = "v27.0.0"
+version = "27.0.2"
 
 [workspace.dependencies.naga]
 package = "naga"
-git = "https://github.com/gfx-rs/wgpu"
-tag = "v27.0.0"
+version = "27.0.0"
 
 [lib]
 crate-type = ["cdylib", "staticlib"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,13 +40,7 @@ version = "27.0.0"
 crate-type = ["cdylib", "staticlib"]
 
 [features]
-default = ["wgsl", "spirv", "glsl", "dx12", "metal"]
-
-#! ### Backends
-# --------------------------------------------------------------------
-#! ⚠️ WIP: Not all backends can be manually configured today.
-#! On Windows, Linux & Android the Vulkan & GLES backends are always enabled.
-#! See [#3514](https://github.com/gfx-rs/wgpu/issues/3514) for more details.
+default = ["wgsl", "spirv", "glsl", "dx12", "metal", "vulkan", "gles"]
 
 ## Enables the DX12 backend on Windows.
 dx12 = ["wgc/dx12"]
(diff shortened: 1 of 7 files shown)


melody-rs:
@gfx-rs?

radgeRayden:
Hopefully now that webgpu.h is stable this kind of major version upgrade will be more trivial to get going.

melody-rs:
It's not that difficult, it's moreso that wgpu-native is lacking contributors

src/lib.rs
```diff
@@ -182,7 +187,8 @@ impl Drop for WGPUDeviceImpl {
         if !thread::panicking() {
             let context = &self.context;
 
-            match context.device_poll(self.id, wgt::PollType::Wait) {
+            // wait_indefinitely() *should* match the old behavior of using wgt::PollType::Wait
```
cwfitzgerald: This new behavior is fine, but note previously there was a 60s timeout.

src/lib.rs
```diff
@@ -2852,8 +2850,13 @@ pub unsafe extern "C" fn wgpuPipelineLayoutRelease(pipeline_layout: native::WGPU
...
 #[no_mangle]
-pub unsafe extern "C" fn wgpuQuerySetDestroy(_query_set: native::WGPUQuerySet) {
-    //TODO: needs to be implemented in wgpu-core
+pub unsafe extern "C" fn wgpuQuerySetDestroy(query_set: native::WGPUQuerySet) {
+    let (query_set_id, context) = {
+        let query_set = query_set.as_ref().expect("query set");
+        (query_set.id, &query_set.context)
+    };
+
+    context.query_set_drop(query_set_id);
```
cwfitzgerald, line 2860: We should file an issue - this is using `drop` to implement destroy. Better than nothing, but we shouldn't lose track of this needing to be destroy.

Cargo.toml
```diff
@@ -23,22 +23,22 @@ resolver = "2"
 [workspace.dependencies.wgc]
 package = "wgpu-core"
 git = "https://github.com/gfx-rs/wgpu"
-tag = "v25.0.2"
+tag = "v27.0.0"
```
cwfitzgerald: Now that there's some patches out, lets bump this to v27.0.2

cwfitzgerald (changes requested):
Sorry for the delay, some comments, but then we can land this!

Cargo.toml
melody-rs: Just noticed that it uses git dependencies, should I change it to use wgpu from crates.io instead?

src/lib.rs
melody-rs: I'm reading through the spec to figure out what the destroy behavior is actually supposed to be, and noticed that `wgpuDeviceRelease`'s behavior was changed recently:

> Releasing a Device object {#DeviceRelease}
> 
> Unlike other destroyable objects, releasing the last (external) ref to a device causes it to be automatically destroyed, if it isn't already lost or destroyed. Though the device object is no longer valid to use after releasing the last ref, the direct effects of @ref wgpuDeviceDestroy() still take effect:
> 
>     It destroys all buffers on the device, which will unmap them and abort any pending map requests.
>     It loses the device and triggers the DeviceLost event.
> 
> Because the device's last ref has already been released, DeviceLost callbacks triggered in this way will not receive a pointer to the device object. More specifically, freeing the last ref:
> 
>     Sets a flag on the DeviceLost future (if it is uncompleted, even if it is already ready) indicating it should pass a null @ref WGPUDevice to the callback.
>     Calls @ref wgpuDeviceDestroy(), readying the DeviceLost future.
>         This may call the DeviceLost callback, if it is @ref WGPUCallbackMode_AllowSpontaneous.
>     Decrements the refcount, bringing it to 0.

I'll go ahead and make `wgpuDeviceRelease` follow the spec

melody-rs: Actually maybe we're already following the spec?

Cargo.toml
cwfitzgerald: I think having some flexibility here is fine. This is also just how things have always gone.

cwfitzgerald (approved):
Lets land and iterate :)

PR #511, opened 2025-08-30
Upgrade webgpu headers

eliemichel:
The submodule that gets the officiel `webgpu.h` header is 9 months old, and a lot happened since then. This branch is a WIP upgrade to a more recent version.


cwfitzgerald:
Sorry that our reviewing was so delayed, trying to work this into my regular reviewing schedule so that PRs to wgpu-native don't hang for too long, this is on us.

Closing in favor of #536

PR #523, opened 2025-11-08
Implement futures

dotcarmen:
This PR implements `wgpuInstanceWaitAny` and modifies already-implemented functions that rely on this functionality.

Apologies for the large diff. It's mostly contained to the first commit, which refactors `Context` to be a struct that contains `wgc::global::Global` instead of being an alias to that type.

Additionally, this PR somewhat replicates logic that's already implemented in `wgc`, but are hidden from the public api (aka declared using `pub(crate)` and are not re-exported elsewhere). Unfortunately, this includes `wgc::lock`, which I didn't duplicate here - instead, I simply rely on `parking_lot` without incorporating the feature-gated implementations from `wgc`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,8 +21,8 @@ use std::{
     thread,
 };
 use utils::{
-    get_base_device_limits_from_adapter_limits, make_slice, str_into_string_view,
-    string_view_into_label, string_view_into_str, texture_format_has_depth,
+    get_base_device_limits_from_adapter_limits, make_slice, make_slice_mut, str_into_string_view,
+    string_view_into_label, string_view_into_str, texture_format_has_depth, FutureRegistry,
 };
 use wgc::{
     command::{bundle_ffi, ComputePass, RenderPass},
@@ -46,6 +46,7 @@ pub mod native {
 
 pub struct Context {
     global: wgc::global::Global,
+    futures: parking_lot::RwLock<FutureRegistry>,
 }
 
 pub struct WGPUAdapterImpl {
@@ -673,6 +674,7 @@ pub unsafe extern "C" fn wgpuCreateInstance(
     Arc::into_raw(Arc::new(WGPUInstanceImpl {
         context: Arc::new(Context {
             global: wgc::global::Global::new("wgpu", &instance_desc),
+            futures: Default::default(),
         }),
     }))
 }
@@ -865,7 +867,8 @@ pub unsafe extern "C" fn wgpuAdapterRequestDevice(
         }
     };
 
-    NULL_FUTURE
+    // `context.global.adapter_request_device` resolves immediately
+    context.futures.write().completed_future().into()
 }
 
 #[no_mangle]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -44,7 +45,10 @@ pub mod native {
     include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 }
 
-pub type Context = wgc::global::Global;
+pub struct Context {
+    global: wgc::global::Global,
+    futures: parking_lot::RwLock<FutureRegistry>,
+}
```
dotcarmen: i can reduce the diff with `impl Deref`, let me know if this is preferred :)

dotcarmen:
i won't keep force-pushing now that it's open for review - was trying to keep the set of commits small for maintainers' convenience :) i recommend looking at the 2nd and 3rd commits to isolate away the changes from the refactor

src/lib.rs
```diff
@@ -44,7 +44,10 @@ pub mod native {
     include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 }
 
-pub type Context = wgc::global::Global;
+pub struct Context {
+    global: wgc::global::Global,
```
dotcarmen, line 48: i can reduce the diff with impl Deref, let me know if this is preferred :)

src/lib.rs
```diff
@@ -4742,3 +4752,46 @@ pub unsafe extern "C" fn wgpuRenderPassEncoderWriteTimestamp(
...
+        drop(registry);
+
+        if success {
+            return native::WGPUWaitStatus_Success;
+        }
+
+        let now = std::time::Instant::now();
+        if now - start >= std::time::Duration::from_nanos(timeout_ns) {
+            return native::WGPUWaitStatus_TimedOut;
+        }
```
dotcarmen, line 4795: i figured it's simple to keep this as a hot loop than to park the thread, but i can also replace the whole loop with:
```rs
let mut success = false;
{
  let registry = context.futures.read();
  for future in futures.iter_mut() {
    // ... the same
  }
}

if success {
  return native::WGPUWaitStatus_Success;
}

std::thread::sleep(timeout_ns);
let registry = context.futures.read();
for future in futures.iter_mut() {
  // ... the same
}

return if success { native::WGPUWaitStatus_Success } else { native::WGPUWaitStatus_TimedOut };
```

src/utils.rs
```diff
@@ -566,3 +579,101 @@ pub fn test_get_base_device_limits_from_adapter_limits() {
...
+    }
+
+    /// Creates a `FutureId` that's incomplete. Call `complete` to mark the
+    /// future completed.
+    pub fn incomplete_future(&mut self) -> FutureId {
+        let id = self.identity.process();
+        let (idx, epoch) = id.unzip();
+        if idx as usize >= self.futures.len() {
+            self.futures
+                .resize_with(idx as usize + 1, || FutureElement::Vacant);
```
dotcarmen, line 645: eg
```rs
let mut reg = FutureRegistry::default();
for _ in 0..253 {
  reg.completed_future();
}

// still 0 allocations, `reg.futures.len() == 0`
reg.completed_future();
// 1 allocation, `reg.futures.len() == 255`
reg.incomplete_future();
```

dotcarmen: wait no, `reg.futures.len() == 1` in the 2nd scenario since the underlying `IdentityManager` is able to only increment the epoch 😅

dotcarmen:
making this a draft again - i need to make sure to call poll functions appropriately

PR #534, opened 2025-12-01
Implement wgpuQueueGetTimestampPeriod

lisyarus:
As of now, timestamp queries are *technically* unusable since the timestamp period is unknown (although I guess it is quite often simply equal to 1). This pull request fixes that.

It **adds a new function** to `wgpu.h`. I haven't found info on the policy regarding extending the API, so I took the liberty to just add the function, since it's a small addition, and definitely not a breaking change :)


almarklein:
Thanks!

PR #542, opened 2025-12-22
fix: ensure command encoder cleanup always runs

### LINKED ISSUE #527: Memory leak from command encoder on Windows & macOS
I initially discovered this when writing my first webgpu program, and it drove me crazy as I thought I was at fault. It appears that even the most basic program will leak a small amount of memory each frame, including the triangle example. 

With the triangle example it seems to be about 47 MB every 5 minutes which equals 156 kB / second or about a kilobyte per frame (my monitor is 165hz). 

Image

Visual Studio's heap profiler blames wgpuDeviceCreateCommandEncoder for the leaked allocations:

Image

Those screenshots are from profiling triangle.exe from the current main branch, but first noticed this when making my own program, which just encodes a single render pass with no pipeline:
```cpp
int main() {
    // setup GLFW (window library)
    glfwInit();

    // create window
    glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API); // we aren't going to use openGL
    glfwWindowHint(GLFW_RESIZABLE, GLFW_FALSE); // don't make it resizable (complicated)
    GLFWwindow *window = glfwCreateWindow(640, 480, "Hello World", NULL, NULL);

     // create surface, device, adapter, etc
    GpuState state = InitGPU(window);

    while (!glfwWindowShouldClose(window)) {
        glfwPollEvents();

        WGPUSurfaceTexture texture;
        wgpuSurfaceGetCurrentTexture(state.surface, &texture);

        WGPUTextureView view = wgpuTextureCreateView(texture.texture, nullptr);

        WGPURenderPassColorAttachment attachment = {
            .view = view,
            .resolveTarget = nullptr,
            .loadOp = WGPULoadOp_Clear,
            .storeOp = WGPUStoreOp_Store,
            .clearValue = WGPUColor(1, 0, 0, 1),
        };

        WGPUCommandEncoder encoder = wgpuDeviceCreateCommandEncoder(state.device, nullptr);
        WGPURenderPassDescriptor renderPass = {};
        renderPass.colorAttachmentCount = 1;
        renderPass.colorAttachments = &attachment;

        WGPURenderPassEncoder pass = wgpuCommandEncoderBeginRenderPass(encoder, &renderPass);
        wgpuRenderPassEncoderEnd(pass);
        wgpuRenderPassEncoderRelease(pass);

        WGPUCommandBuffer buffer = wgpuCommandEncoderFinish(encoder, nullptr);


        wgpuQueueSubmit(state.queue, 1, &buffer);

        wgpuSurfacePresent(state.surface);

        wgpuCommandBufferRelease(buffer);
        wgpuCommandEncoderRelease(encoder);

        wgpuTextureViewRelease(view);
        wgpuTextureRelease(texture.texture);

        // main loop

    }

    glfwTerminate();
    return 0;
}
```

### LINKED ISSUE #541: Command encoder memory leak on Metal (and all backends) - cleanup never runs
Command encoders leak memory because their Drop cleanup code is skipped. This affects all backends (Metal, Vulkan, DX12) but is most visible on Metal where it causes ~0.5-1 MB leak per 10 seconds at 60fps.

This is related to #527 but provides additional evidence and a minimal reproduction.

### Root Cause

In `wgpuCommandEncoderFinish` (src/lib.rs), the encoder is marked closed:

```rust
command_encoder.open.store(false, atomic::Ordering::SeqCst);
```

But `WGPUCommandEncoderImpl::Drop` only runs cleanup when `open == true`:

```rust
impl Drop for WGPUCommandEncoderImpl {
    fn drop(&mut self) {
        if self.open.load(atomic::Ordering::SeqCst) && !thread::panicking() {
            self.context.command_encoder_drop(self.id);
        }
    }
}
```

Since `finish()` sets `open = false` before `Release()` is called, the `command_encoder_drop()` cleanup **never executes**.

### Evidence: Rust wgpu vs C++ wgpu-native

I ran identical tests - same workload, same Metal backend:

| Implementation | Memory Behavior |
|----------------|-----------------|
| Rust wgpu 24.0 | Stable after warm-up (no leak) |
| C++ wgpu-native v27.0.2.0 | +0.5 MB per 10 seconds (leaks) |

This proves the leak is in wgpu-native's FFI layer, not the Metal driver or wgpu-core.

### Minimal Reproduction

```cpp
#include <webgpu/webgpu.h>
#include <mach/mach.h>
#include <iostream>
#include <thread>
#include <chrono>

double getMemoryMB() {
    mach_task_basic_info info;
    mach_msg_type_number_t count = MACH_TASK_BASIC_INFO_COUNT;
    if (task_info(mach_task_self(), MACH_TASK_BASIC_INFO, (task_info_t)&info, &count) == KERN_SUCCESS)
        return info.resident_size / (1024.0 * 1024.0);
    return 0;
}

int main() {
    // ... adapter/device setup omitted for brevity ...

    WGPUTextureDescriptor texDesc = { .size = {256, 256, 1}, .format = WGPUTextureFormat_BGRA8Unorm,
        .usage = WGPUTextureUsage_RenderAttachment, .mipLevelCount = 1, .sampleCount = 1 };
    WGPUTexture texture = wgpuDeviceCreateTexture(device, &texDesc);
    WGPUTextureView view = wgpuTextureCreateView(texture, nullptr);

    double startMem = getMemoryMB();
    int frame = 0;

    while (true) {
        WGPUCommandEncoder encoder = wgpuDeviceCreateCommandEncoder(device, nullptr);

        WGPURenderPassColorAttachment colorAtt = { .view = view, .loadOp = WGPULoadOp_Clear,
            .storeOp = WGPUStoreOp_Store, .depthSlice = WGPU_DEPTH_SLICE_UNDEFINED };
        WGPURenderPassDescriptor passDesc = { .colorAttachmentCount = 1, .colorAttachments = &colorAtt };

        WGPURenderPassEncoder pass = wgpuCommandEncoderBeginRenderPass(encoder, &passDesc);
        wgpuRenderPassEncoderEnd(pass);
        wgpuRenderPassEncoderRelease(pass);

        WGPUCommandBuffer cmdBuf = wgpuCommandEncoderFinish(encoder, nullptr);  // Sets open=false
        wgpuQueueSubmit(queue, 1, &cmdBuf);
        wgpuCommandBufferRelease(cmdBuf);
        wgpuCommandEncoderRelease(encoder);  // Drop runs but cleanup is skipped!

        if (++frame % 600 == 0) {
            double mem = getMemoryMB();
            std::cout << "Memory: " << (int)mem << " MB (leaked: +" << (int)(mem - startMem) << " MB)\n";
        }
        std::this_thread::sleep_for(std::chrono::milliseconds(16));
    }
}
```

### Proposed Fix

The `open` flag appears to guard against double-drop when an encoder is dropped without being finished. The fix should ensure cleanup runs in both cases:

**Option A:** Remove the `open` check from Drop entirely, track "needs cleanup" separately
**Option B:** Don't set `open = false` in `finish()`, only set it in Drop after cleanup runs
**Option C:** Call `command_encoder_drop()` directly in `finish()` instead of relying on Drop

I'm happy to submit a PR with whichever approach maintainers prefer.

### Environment

- macOS Sequoia 15.2, Apple M4 Max
- wgpu-native v27.0.2.0
- Also reproduced on macOS Sonoma (M1) and macOS Ventura (Intel)

### Related Issues

- #527 - Same root cause, different reproduction
- gfx-rs/wgpu#8768 - Original report (before root cause was identified)

jeffcrouse:
### Summary

The Drop implementation for `WGPUCommandEncoderImpl` only ran cleanup when `self.open` was true, but `wgpuCommandEncoderFinish` set `open=false` before the encoder was released. This meant `command_encoder_drop()` was never called, causing memory to leak.

### The Bug

```rust
// In wgpuCommandEncoderFinish (line 1401):
command_encoder.open.store(false, atomic::Ordering::SeqCst);

// In Drop (line 129):
if self.open.load(atomic::Ordering::SeqCst) && !thread::panicking() {
    context.command_encoder_drop(self.id);  // Never reached!
}
```

### The Fix

Remove the `open` check from Drop - cleanup should always run when the encoder is released:

```rust
if !thread::panicking() {
    context.command_encoder_drop(self.id);
}
```

### Test Results

| Version | Memory at 0s | Memory at 180s | Behavior |
|---------|--------------|----------------|----------|
| Before fix | 23 MB | 33 MB | +0.5 MB every 10 seconds |
| After fix | 21 MB | 21 MB | Stable |

### Reproduction

Minimal C++ test that creates command encoders at 60fps:

```cpp
while (true) {
    WGPUCommandEncoder encoder = wgpuDeviceCreateCommandEncoder(device, nullptr);
    WGPURenderPassEncoder pass = wgpuCommandEncoderBeginRenderPass(encoder, &passDesc);
    wgpuRenderPassEncoderEnd(pass);
    wgpuRenderPassEncoderRelease(pass);
    WGPUCommandBuffer cmdBuf = wgpuCommandEncoderFinish(encoder, nullptr);
    wgpuQueueSubmit(queue, 1, &cmdBuf);
    wgpuCommandBufferRelease(cmdBuf);
    wgpuCommandEncoderRelease(encoder);
    std::this_thread::sleep_for(std::chrono::milliseconds(16));
}
```

Fixes #541
Fixes #527

### Notes

- The `open` flag still serves a purpose in `wgpuCommandEncoderFinish` to prevent double-finishing
- This fix ensures cleanup happens regardless of whether `finish()` was called
- Tested on macOS Sequoia / Apple M4 Max with wgpu-native v27

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,7 +126,7 @@ pub struct WGPUCommandEncoderImpl {
 }
 impl Drop for WGPUCommandEncoderImpl {
     fn drop(&mut self) {
-        if self.open.load(atomic::Ordering::SeqCst) && !thread::panicking() {
+        if !thread::panicking() {
             let context = &self.context;
             context.command_encoder_drop(self.id);
         }


cwfitzgerald (approved):
Nice

PR #197, opened 2022-06-19
Implement GLSL shader source

MineGame159:
Implemented GLSL shader source with support for passing in pre-processor defines.
From my testing it works perfectly.
Closes #195.

PR #545, opened 2026-01-01
make `WGPUShaderSourceGLSL::defines` const

xor-shift:
The only use of this field i could find was in `conv.rs` and it was used with `make_slice` [here](https://github.com/gfx-rs/wgpu-native/blob/ba4deb5d935652f40c7e051b15cbb5d097219941/src/conv.rs#L663) which expects an immutable pointer. I tried finding justification for this field being mutable but the field was first used [here](https://github.com/gfx-rs/wgpu-native/pull/197/files#diff-a9d8c4b4b4059ecb0003b87ef8931eb21aef7a9cec0b504a0d4f13579ae70b5cR410), as an immutable pointer.


cwfitzgerald (approved):
Thank you!

PR #550, opened 2026-02-08
Expose graphics debugger capture functions in wgpu.h

ShadelessFox:
Quite useful when rendering to an off-screen buffer.

I would also like to extend the wgpu API to report whether a graphics debugger capture is **available** (and, if not, return a string indicating the reason, similar to `RenderDoc::NotAvailable`) and whether it's currently **active**, as it's currently up to the library's consumer to track; both RenderDoc and Metal expose such APIs, so it should be pretty straightforward to implement. Please let me know what you think.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2497,6 +2497,30 @@ pub unsafe extern "C" fn wgpuDeviceGetFeatures(
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn wgpuDeviceStartGraphicsDebuggerCapture(
+    device: native::WGPUDevice
+) -> bool {
+    let (device_id, context) = {
+        let device = device.as_ref().expect("invalid device");
+        (device.id, &device.context)
+    };
+
+    // FIXME: wgpu-hal's start_graphics_debugger_capture returns a bool, but wgpu's doesn't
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -2496,6 +2496,30 @@ pub unsafe extern "C" fn wgpuDeviceGetFeatures(
...
+#[no_mangle]
+pub unsafe extern "C" fn wgpuDeviceStartGraphicsDebuggerCapture(
+    device: native::WGPUDevice
+) -> bool {
+    let (device_id, context) = {
+        let device = device.as_ref().expect("invalid device");
+        (device.id, &device.context)
+    };
+
+    // FIXME: wgpu-hal's start_graphics_debugger_capture returns a bool, but wgpu's doesn't
```
ShadelessFox, line 2507: I made this function return a boolean regardless, so when wgpu's implementation is updated to return a bool, it can be updated here as well without breaking the ABI.

cwfitzgerald: Sounds good.

cwfitzgerald (approved):
Looks groovy to me!

cwfitzgerald:
CI mad

cwfitzgerald (approved):
Thank you!

PR #557, opened 2026-03-05
metal: expose native device/queue/texture interop handles

jeffcrouse:
## Summary
- add minimal Metal-only native interop getters to the C API:
  - `wgpuDeviceGetNativeMetalDevice`
  - `wgpuQueueGetNativeMetalCommandQueue`
  - `wgpuTextureGetNativeMetalTexture`
- return borrowed opaque pointers (`void*`) to `id<MTLDevice>`, `id<MTLCommandQueue>`, and `id<MTLTexture>`
- return `NULL` when Metal backend handles are unavailable

## Notes
- no behavior changes to existing API paths
- symbols are exported from both static and shared builds via `#[no_mangle] extern "C"`
- header docs describe ownership/lifetime expectations (borrowed handles; caller must not release)

## Validation
- `cargo build` succeeds on macOS
- symbols are present in `libwgpu_native.dylib` via `nm`


cwfitzgerald (approved):
LGTM, thank you!

cwfitzgerald:
Weird, something is up after merging with main

cwfitzgerald:
Could you possibly take a look to see if you can fix the CI issues, this isn't a very old PR so I'm suprised that merging main caused it to fail, otherwise I'll take a look in a few days.

jeffcrouse:
I took a look. Most of the CI failures are cascading from a single `-D warnings` issue on my branch: an unused `core::slice` import in `src/lib.rs`, which is tripping all the Clippy and examples jobs. I just pushed a small fix for that now (`a9a59af`). Once CI reruns, I’ll re-check the remaining `release - ios-aarch64` failure separately if it still reproduces.

jeffcrouse:
I took another pass at the CI break and pushed two follow-ups:

- `a9a59af` removes the unused `core::slice` import that was tripping the broad `-D warnings` failures
- `fc5fe5c` fixes the Apple bindgen setup in `build.rs` (correct SDK/clang target for device vs simulator), adds the relevant `rerun-if-env-changed` hooks, and bumps the CI/CD cache suffix once to flush stale generated bindings on the iOS lanes

That seems to have resolved the remaining iOS breakage:
- `Clippy iOS stable` is green now
- `release - ios-aarch64` is green
- `release - ios-aarch64-simulator` is green
- `release - ios-x86_64-simulator` is green

At this point the only job still running on the current rerun is `release - linux-aarch64`.

PR #536, opened 2025-12-19
Modified Webgpu to work with the stable webgpu-header.

jalexcole:
I updated the header file to latest, and patched the lib and wgpu.h file in order to pass cargo test.

PR #572, opened 2026-04-08
Modified Webgpu to work with the stable webgpu-header.

Murmele:
Fixed the ci issues from and closes https://github.com/gfx-rs/wgpu-native/pull/536


cwfitzgerald:
Thank you!

Murmele:
Thanks for merging!

kolkov:
Great work @Murmele — this covered everything we flagged in our earlier review of #536. Thanks for getting this across the finish line.

Two small observations (if I'm reading the code correctly):

1. **`requiredFeatures` in `wgpuCreateInstance`** — `map_instance_descriptor` in `conv.rs` seems to ignore the base descriptor (`_base` parameter unused), so `requiredFeatures`/`requiredFeatureCount` are never validated. A user passing `WGPUInstanceFeatureName_TimedWaitAny` wouldn't get an error even though TimedWaitAny is unsupported.

2. **Stubs with `unimplemented!()`** — `wgpuGetInstanceFeatures`, `wgpuHasInstanceFeature`, `wgpuBufferReadMappedRange`, `wgpuBufferWriteMappedRange` will panic if called. Downstream bindings that probe capabilities at startup may hit these unexpectedly. Returning empty results or `WGPUStatus_Error` would be safer.

Neither is blocking — just flagging for awareness. Happy to submit a follow-up PR if useful.

PR #581, opened 2026-04-16
Implement ClampToBorder (fixes #521)

### LINKED ISSUE #521: Why ClampToBorder and ClampToZero are disabled?
Hi, thank you for your work and effort in maintaining wgpu-native. My question is, why are **ClampToBorder** and **ClampToZero** still [disabled](https://github.com/gfx-rs/wgpu-native/blob/b1ff7f6ef6fa274327855050896f8bb77770e358/ffi/wgpu.h#L39)?

    // TODO: requires wgpu.h api change
    // WGPUNativeFeature_AddressModeClampToZero = 0x00030011,
    // WGPUNativeFeature_AddressModeClampToBorder = 0x00030012,

lisyarus:
## Checklist

## Description

Uncommencted the `ClampToBorder` and `ClampToZero` features. Since these aren't present in `webgpu.h`, I had to add a few things to `wgpu.h` API:

* Add `WGPUNativeAddressMode` enum with `WGPUNativeAddressMode_ClampToBorder` value, which is supposed to be used in places where `WGPUAddressMode` is expected
* Add `WGPUSamplerBorderColor` enum that mimicks the [corresponding wgpu enum](https://wgpu.rs/doc/wgpu_types/texture/enum.SamplerBorderColor.html)
* Add `WGPUSamplerDescriptorExtras` struct to be chained in `WGPUSamplerDescriptor`, which provides the `WGPUSamplerBorderColor` value.

I had to rewrite the `map_address_mode` function in `conv.rs`: preciously it was generated using the `map_enum_with_undefined` macro, but since now it mixes values from two native enums, I had to rewrite it by hand.

Currently, if `WGPUNativeAddressMode_ClampToBorder` is used but the `WGPUSamplerDescriptorExtras` isn't provided, it sets the `borderColor` in [`SamplerDescriptor`](https://docs.rs/wgpu/latest/wgpu/type.SamplerDescriptor.html) to `None`, and the wgpu itself is supposed to handle this case. We could alternatively check that explicitly in `wgpuDeviceCreateSampler`, I'm not sure what the correct thing to do would be here.

## Related Issues

This PR fixes #521

### DIFF
--- a/src/conv.rs
+++ b/src/conv.rs
@@ -39,15 +39,6 @@ map_enum_with_undefined!(
     Store
 );
 map_enum_with_undefined!(
-    map_address_mode,
-    WGPUAddressMode,
-    wgt::AddressMode,
-    "Unknown address mode",
-    ClampToEdge,
-    Repeat,
-    MirrorRepeat
-);
-map_enum_with_undefined!(
     map_filter_mode,
     WGPUFilterMode,
     wgt::FilterMode,
@@ -249,6 +240,17 @@ map_enum_with_undefined!(
     ReadWrite
 );
 
+map_enum_with_undefined!(
+    map_sampler_border_color,
+    WGPUSamplerBorderColor,
+    wgt::SamplerBorderColor,
+    "Unknown sampler border color",
+    TransparentBlack,
+    OpaqueBlack,
+    OpaqueWhite,
+    Zero
+);
+
 // These are defined as UINT64_MAX in the header, but bindgen currently can't process that define.
 // See https://github.com/rust-lang/rust-bindgen/issues/2822
 pub const WGPU_WHOLE_SIZE: u64 = u64::MAX;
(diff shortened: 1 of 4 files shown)


src/conv.rs
```diff
@@ -38,15 +38,6 @@ map_enum_with_undefined!(
...
 );
-map_enum_with_undefined!(
-    map_address_mode,
-    WGPUAddressMode,
-    wgt::AddressMode,
-    "Unknown address mode",
-    ClampToEdge,
-    Repeat,
-    MirrorRepeat
-);
```
cwfitzgerald, line 49: Question: why did this need to be converted to an explicit function?

cwfitzgerald (approved):
Got a question, and this needs rebasing but looks good.

src/conv.rs
lisyarus: If I understand it correctly, map_enum_with_undefined appends the native enum name (to get WGPUAddressMode_Repeat and the like), but now it has to mix enum values from two different enums (WGPUAddressMode and WGPUNativeAddressMode).

lisyarus:
Rebased to fresh trunk.

cwfitzgerald (approved):

cwfitzgerald:
Rebased

PR #575, opened 2026-04-11
Update native limits (fixes #507)

lisyarus:
Updated the WGPUNativeLimits struct to reflect the changes in wgpu-core and webgpu-headers. Specifically,

* Removed `maxImmediateSize`, as it is now part of `WGPULimits`
* Added `maxBindingArraySamplerElementsPerShaderStage` and `maxMultiviewViewCount`

I've omitted the limits related to mesh shaders & ray query acceleration structres, as afaik the corresponding API is currently not exposed in wgpu-native anyway.

I've also fixed the `immediates` and `texture_arrays` examples. The latter one is what [sparked this PR](https://github.com/gfx-rs/wgpu-native/issues/507) in the first place; it had to be reworked a bit, as texture arrays cannot be present in the same bind group as buffers with dynamic offsets, thus I've split the bound group into two.

I've took the liberty of introducing a `WGPU_NATIVE_LIMITS_INIT` macro that mimicks the `WGPU_LIMITS_INIT` macro from `webgpu.h`.

## Checklist

- [x] `cargo clippy` reports no issues
- [x] `cargo doc` reports no issues
- [x] [`cargo deny`](https://github.com/EmbarkStudios/cargo-deny/) issues have been fixed or added to `deny.toml`
- [x] Human-readable change descriptions added to CHANGELOG.md under the "Unreleased" heading.
  - [ ] If the change does not affect the user (or is a process change), preface the change with "Internal:"
  - [x] Add credit to yourself for each change: `Added new functionality. @githubname`

PR #583, opened 2026-04-16
Even easier Immediates

Vipitis:
## Checklist

- [?] `cargo doc` reports no issues

## Description
While doing the v29 update for the python wrapper. I spotted that the descriptor already has a `immediateSize` field in `webgpu.h` https://github.com/webgpu-native/webgpu-headers/pull/529
Also saw the limits being simplified in #575 so this makes it even simpler, although you still need the native only feature. I updated the example and it still works, please let me know if the changes to the `wgpu.h` headers are fine like this.

### DIFF
--- a/ffi/wgpu.h
+++ b/ffi/wgpu.h
@@ -22,23 +22,23 @@ typedef enum WGPUNativeSType
     /** Identifies @ref WGPUNativeLimits. */
     WGPUSType_NativeLimits = 0x00030002,
     /** Identifies @ref WGPUShaderSourceGLSL. */
-    WGPUSType_ShaderSourceGLSL = 0x00030004,
+    WGPUSType_ShaderSourceGLSL = 0x00030003,
     /** Identifies @ref WGPUInstanceExtras. */
-    WGPUSType_InstanceExtras = 0x00030006,
+    WGPUSType_InstanceExtras = 0x00030004,
     /** Identifies @ref WGPUBindGroupEntryExtras. */
-    WGPUSType_BindGroupEntryExtras = 0x00030007,
+    WGPUSType_BindGroupEntryExtras = 0x00030005,
     /** Identifies @ref WGPUBindGroupLayoutEntryExtras. */
-    WGPUSType_BindGroupLayoutEntryExtras = 0x00030008,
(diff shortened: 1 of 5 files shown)


lisyarus:
Nice! I think the changelog diff is a bit incorrect: it marks as released in v29 everything currently marked as unreleased, which isn't the case (there were new additions after v29). Apparently the current changelog in trunk contains both pre-v29 and post-v29 changes. Maybe just add your changes to unreleased for now, and then a separate PR would sort out the changelog itself?
@cwfitzgerald

Vipitis:
yeah, I quickly looked at the blame and all your recentt changes should probably move to the top as well. I just tried to have mine a little more correct for now. I can add a commit to move yours tomorrow, or maybe do it outside this PR?

lisyarus:
I'd suggest fixing the changelog in a separate PR, just to keep the history clean?

ffi/wgpu.h
```diff
@@ -21,8 +21,6 @@ typedef enum WGPUNativeSType
     WGPUSType_DeviceExtras = 0x00030001,
     /** Identifies @ref WGPUNativeLimits. */
     WGPUSType_NativeLimits = 0x00030002,
-    /** Identifies @ref WGPUPipelineLayoutExtras. */
-    WGPUSType_PipelineLayoutExtras = 0x00030003,
```
Vipitis, line 25: as this enum is removed, do all the others need to move up?

cwfitzgerald: /shrug Probably as we don't really care about backwards/forward compat

cwfitzgerald (approved):
Music to my eyes!

Vipitis:
hey @cwfitzgerald I updated the enums (although there are other enums with gaps still remain, unrealted to the change here).

plus there is even even easier immediates for the set functions that got merged into the webgpu.h header: https://github.com/webgpu-native/webgpu-headers/pull/588 so I will look into that PR next.

cwfitzgerald:
Could you rebase this, sorry for the long delay!

Vipitis:
rebased to newest trunk

cwfitzgerald:
Thanks!

PR #600, opened 2026-06-18
remove duplicate native texture formats

### LINKED ISSUE #596: native texture formats now in webgpu.h should be removed
I run into an issue trying out R16Unorm leading me to `Unknown texture format`. I was using the new enum from `webgpu.h` which is `0x00000005` and that was odd. However the enum from `wgpu.h` (`0x00030001`) did work. Our mapping just took the first one... which I might patch for this release - but I think this is confusing.

So I propose we remove these texture formats from native and just use the spec version?

also came across the feature `WGPUNativeFeature_TextureFormat16bitNorm` -> can we drop this? or is the `webgpu.h` ahead of the wgpu-core release?

Vipitis:
closes #596 

since this is kinda the headers being ahead of v29 (I think tiers are added in v30) this still needs the native only feature (https://github.com/gfx-rs/wgpu/issues/8122)
I changed the example just to test it for myself.. would need a bit of fancying keep it this way

## Checklist

## Description

a question: can we also render to a 16unorm texture?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ The format is loosely based on [Keep a Changelog](https://keepachangelog.com/en/
 
 ## Unreleased
 
+- moved 16bit norm textures into spec and out of wgpu.h `WGPUNativeTextureFormat_Rgba16Unorm` -> `WGPUTextureFormat_RGBA16Unorm`. by @Vipitis in [#tbd](tbd)
+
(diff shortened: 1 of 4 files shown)


almarklein:
Looks good! Let's either revert the example or clean it up?

cwfitzgerald (approved):
Nice!

PR #602, opened 2026-06-20
fix `WGPU_WHOLE_MAP_SIZE` case in `bufferMapAsync`

### LINKED ISSUE #601: wgpuBufferMapAsync fails when size is WGPU_WHOLE_MAP_SIZE
Using version 29.0.0.0.

I am trying to map a buffer using wgpuBufferMapAsync. The doc comment in webgpu.h looks like this:

```cpp
/**
 * @param mode
 * The mapping mode (read or write).
 *
 * @param offset
 * Byte offset relative to beginning of the buffer.
 *
 * @param size
 * Byte size of the region to map.
 * If this is @ref WGPU_WHOLE_MAP_SIZE, it defaults to `buffer.size - offset`.
 */
WGPU_EXPORT WGPUFuture wgpuBufferMapAsync(WGPUBuffer buffer, WGPUMapMode mode, size_t offset, size_t size, WGPUBufferMapCallbackInfo callbackInfo) WGPU_FUNCTION_ATTRIBUTE;
```

When I try and map a buffer with `size == WGPU_WHOLE_MAP_SIZE`, I get the following validation error:

```
Validation Error

Caused by:
  Buffer range size invalid: range_size 18446744073709551615 must be multiple of 4
```

Since the doc comment specifically calls out using WGPU_WHOLE_MAP_SIZE as the size, I assume this is a bug.

Vipitis:
fixes #601 

## Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1064,7 +1064,10 @@ pub unsafe extern "C" fn wgpuBufferMapAsync(
     if let Err(cause) = context.buffer_map_async(
         buffer_id,
         offset as wgt::BufferAddress,
-        Some(size as wgt::BufferAddress),
+        match size {
+            conv::WGPU_WHOLE_MAP_SIZE => None,
+            _ => Some(size as wgt::BufferAddress),
+        },
         operation,
     ) {
         handle_error(error_sink, cause, None, "wgpuBufferMapAsync");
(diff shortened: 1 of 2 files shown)


src/lib.rs
Vipitis, line 997: question for the rust people: is there any reason not use the use type alias (?) here too. `wgt::BufferAddress` instead of `u64`

cwfitzgerald (approved):

src/lib.rs
cwfitzgerald: Formally, yes this should be wgt::BufferAddress, but also meh

PR #606, opened 2026-06-26
Wgpu yaml

jalexcole:
I have created a yaml file in the style of webgpu.yml. The purpose of this is to aid code generation for bindings.

I did have a deviation which I called `extension-objects` that extend the functionality of the webgpu.yml objects. I would like in the future if the wgpu.h could be built off of this and allow for comments to be added.


Vipitis:
This will be awesome to have! Would it make sense to use the generator tooling from webgpu-headers to produce the header based on the yaml?

jalexcole:
@Vipitis For utilizing the same script, it could be done but it would also need to include webgpu.h. I also think it would have to then be split into two separate yaml files for enum values since wgpu is inconsistent on starting from 0 and starting from 0x30000.

Vipitis:
@jalexcole I haven't tried it or looked at the source. But it does sound the upstream script is meant to support such a purpose: https://github.com/webgpu-native/webgpu-headers/tree/main/gen#generate-implementation-specific-header webgpu.h should stay consistent with the tracked submodule commit.

We would have the benefit of knowing that our `wgpu-` yaml (, json) and header are in sync. Makes it easier to maintain and trust too when there is a CI job that runs it to make sure all changes are committed.

cwfitzgerald:
Lets merge this as is, then could you file an issue/PR for moving over to using the proper script?

PR #616, opened 2026-07-08
CD: Drop qemu in favor of arm linux runner

almarklein:
Linux arm arm runners are available in public repos see [gh blog](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/) since about a year and a half. Also see [GH's list of runners](https://docs.github.com/en/actions/reference/runners/github-hosted-runners).

This means we can drop QEMU. The build time for linux-aarch64 goes down from 45 min to 4 min like the rest. 🥳

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -212,15 +212,17 @@ jobs:
   linux:
     # Config
     name: release - ${{ matrix.name }}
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
That's amazing!

PR #594, opened 2026-05-26
Complete parity with wgpu

inner-daemons:
Closes #553 

## Checklist

- [x] `cargo clippy` reports no issues
- [x] `cargo doc` reports no issues
- [x] [`cargo deny`](https://github.com/EmbarkStudios/cargo-deny/) issues have been fixed or added to `deny.toml`
- [x] Human-readable change descriptions added to CHANGELOG.md under the "Unreleased" heading.
  - [x] If the change does not affect the user (or is a process change), preface the change with "Internal:"
  - [x] Add credit to yourself for each change: `Added new functionality. @githubname`

## Description
This is this sister PR to https://github.com/gfx-rs/wgpu/pull/9570. It brings wgpu-native into full parity with wgpu such that all features are supported, and all tests pass.

I cannot guarantee that everything is 100% right but I plan to look over every change and claude has also looked over every change, and like I mentioned it passes the full test suite of wgpu.

The only real breaking change is the removal of `wgpuDeviceCreateShaderModuleSpirV` which has been replaced with wgpu's new passthrough API. It is easily replaced.

PR #617, opened 2026-07-10
Complete parity with wgpu 30.0, part 1

inner-daemons:
## Checklist

## Description
Broken off from #594. See discussion there.

This brings basically all functionality from wpgu into wgpu-native, and has a few fixes for broken existing functionality. It was tested but that won't be part of this PR and will come later.


ffi/wgpu.h
```diff
@@ -35,10 +39,30 @@ typedef enum WGPUNativeSType
     WGPUSType_SurfaceConfigurationExtras = 0x00030008,
     /** Identifies @ref WGPUSurfaceSourceSwapChainPanel. */
     WGPUSType_SurfaceSourceSwapChainPanel = 0x00030009,
-    /** Identifies @ref WGPUPrimitiveStateExtras. */
     WGPUSType_PrimitiveStateExtras = 0x0003000A,
     /** Identifies @ref WGPUSamplerDescriptorExtras. */
     WGPUSType_SamplerDescriptorExtras = 0x0003000B,
+    /** Identifies @ref WGPUComputePipelineDescriptorExtras. */
```
Vipitis, line 45: enum skips `0x0003000C` and `0x00030011`

ffi/wgpu.h
```diff
@@ -335,7 +359,17 @@ typedef enum WGPUNativeFeature
      * This is a native only feature.
      */
     WGPUNativeFeature_StorageTextureArrayNonUniformIndexing = 0x00030010,
+    /**
```
Vipitis, line 362: looks like it's the same upstream, but you need to enable a featured called `AddressMode_ClampToZero` (and also `AddeessMode_ClampToBorder?)` to now use the BorderColor extra with color zero? 

I guess clamp to zero existed before and then other border colors got added so it got generalized away. 

If that's the only way to use it, perhaps a docstring (or reference to example/test) could help. 

E: looks like it was the other way around: https://github.com/gfx-rs/wgpu/pull/2364 still odd to me and something I want to try, even if unrelated to this PR.

ffi/wgpu.h
```diff
@@ -636,12 +683,31 @@ typedef enum WGPUNativeFeature
     WGPUNativeFeature_TextureInt64Atomic = 0x00030030,
     // TODO: not implemented yet, see https://github.com/gfx-rs/wgpu/issues/7149
     // WGPUNativeFeature_UniformBufferBindingArrays = 0x00030031,
-    // TODO: requires wgpu.h api change
-    // WGPUNativeFeature_MeshShader = 0x00030032,
-    // WGPUNativeFeature_RayHitVertexReturn = 0x00030033,
-    // WGPUNativeFeature_MeshShaderMultiview = 0x00030034,
-    // WGPUNativeFeature_ExtendedAccelerationStructureVertexFormats = 0x00030035,
-    // WGPUNativeFeature_PassthroughShaders = 0x00030036,
+    WGPUNativeFeature_MeshShader = 0x00030032,
```
Vipitis, line 686: no docstring on the first one

ffi/wgpu.h
```diff
@@ -810,6 +948,39 @@ static const WGPUInstanceBackend WGPUInstanceBackend_Primary = (1 << 0) | (1 <<
...
+ *
+ * The render target has undefined contents at the start of the render pass.
+ * This is the fastest option when every pixel is overwritten by the pass, but
+ * reading an unwritten pixel is undefined behavior. Backends that don't support
+ * it internally fall back to an unspecified load op. Under
+ * @ref WGPUInstanceFlag_StrictWebgpuCompliance it is rejected.
+ *
+ * Assignable to any @ref WGPULoadOp field (e.g. @ref WGPURenderPassColorAttachment::loadOp).
+ */
+#define WGPULoadOp_DontCare 0x00030001
```
Vipitis, line 962: I don't believe these values should be done via a macro. It should likely be a `WGPUNativeLoadOp`

similar with `WGPUTextureAspect` further down

ffi/wgpu.h
```diff
@@ -1217,6 +1420,33 @@ typedef struct WGPUNativeLimits
      * The maximum number of views that can be used in multiview rendering.
      */
     uint32_t maxMultiviewViewCount;
+    /**
```
Vipitis, line 1423: if these are limits, why are they `InstanceExtras` instead of `NativeLimits`?

ffi/wgpu.h
```diff
@@ -750,10 +820,78 @@ typedef enum WGPUNativeFeature
...
+     *
+     * This is a native only feature.
+     */
+    WGPUNativeFeature_VulkanExternalMemoryDmaBuf = 0x00030042,
+    /**
+     * Enables ray tracing pipelines (@c wgpuRayTracingPipeline*).
+     *
+     * This is an experimental, native-only feature.
+     */
+    WGPUNativeFeature_ExperimentalRayTracingPipelines = 0x00030043,
```
Vipitis, line 840: naming of `Experimental*`, as well as the docstring is not consistent. I think it's fine to name all experimental features experimental - as you also need to enable that for the adapter.

ffi/wgpu.h
```diff
@@ -1350,8 +1681,110 @@ typedef struct WGPUSurfaceConfigurationExtras
...
+    float potential;
+    float reference;
+} WGPUDisplayHeadroom WGPU_STRUCTURE_ATTRIBUTE;
+
+/** CIE 1931 xy chromaticity of the display primaries and white point. Each
+ *  x/y component is @c NaN when unknown. */
+typedef struct WGPUDisplayChromaticity
+{
+    WGPUBool present;
+    float redX, redY;
```
Vipitis, line 1724: I would prefer if they were one per line.

Vipitis (commented):
started reading the diff, some notes... will take me a few more days to get through it and then trying some parts of it too.

inner-daemons:
I will look at these comments and I appreciate them. But I feel a little guilty because I have barely glanced at the actual header changes, you have done more in that regard.

ffi/wgpu.h
```diff
@@ -1397,6 +1880,366 @@ typedef struct WGPUPrimitiveStateExtras
...
+    WGPUAccelerationStructureGeometryFlags flags;
+} WGPUBlasAABBGeometrySizeDescriptor WGPU_STRUCTURE_ATTRIBUTE;
+
+/**
+ * Size descriptors for a BLAS.
+ * Set @ref kind and fill the matching pair of fields.
+ * The other pair should be NULL/0.
+ * Pass to @ref wgpuDeviceCreateBlas.
+ */
+typedef struct WGPUBlasSizeDescriptors
```
Vipitis, line 2106: I was curious on how this kind of union could be solved. Would it make sense to combined the sizes? Also do we need the nullable prefix? Similar example from webgpu.h is `GPUBindGroupEntry`

ffi/wgpu.h
```diff
@@ -1452,6 +2347,144 @@ static const WGPUShaderRuntimeChecks WGPUShaderRuntimeChecks_TaskShaderDispatchT
...
+ */
+static const WGPUShaderRuntimeChecks WGPUShaderRuntimeChecks_IntDivChecks = 0x0000000000000020;
+
+/**
+ * Bitmask of texture format feature flags returned by
+ * @ref wgpuAdapterGetTextureFormatCapabilities.
+ *
+ * Bit values match those of @c wgpu_types::TextureFormatFeatureFlags.
+ */
+typedef uint32_t WGPUNativeTextureFormatFeatureFlags;
```
Vipitis, line 2362: There is a 64bit flag type, probably best to use that.

ffi/wgpu.h
```diff
@@ -1544,6 +2598,143 @@ extern "C"
...
+    /** Returns true if the BLAS is ready to be compacted via @ref wgpuQueueCompactBlas. */
+    WGPUBool wgpuBlasReadyForCompaction(WGPUBlas blas);
+
+    /**
+     * Compact a BLAS. Returns a new (smaller) BLAS handle.
+     * The old handle remains valid and must be released separately.
+     */
+    WGPUBlas wgpuQueueCompactBlas(WGPUQueue queue, WGPUBlas blas);
+
+    /** Query internal wgpu-core/HAL resource counters for debugging. */
```
Vipitis, line 2655: The struct don't include core counters, but it also looks like that is cute and empty placeholder?

cwfitzgerald (changes requested):
There's a number of issues here, I have attached the claude review with a number of them.

<details>
<summary>Details</summary>

# PR #617 — Complete parity with wgpu 30.0, part 1

**Preamble.** Bumps all wgpu crates from v29 to a post-30.0 git rev (`1a3a4cc2`, 2026-07-27), then fills in the FFI surface: ray tracing (BLAS/TLAS, compaction), mesh/task pipelines, pipeline cache, external textures, passthrough shaders, cooperative matrices, HDR/display info, allocator+internal counters, 21 new native limits, 14 ASTC-HDR formats. Structurally: `ffi/wgpu.h` grows +1213 lines of extension declarations, `src/conv.rs` +696 of mapping tables, `src/lib.rs` +1912 of entry points, and `src/unimplemented.rs` loses 29 stubs. No tests are added. Bundle reports **CONFLICTING** against `trunk`.

The mapping tables are, on the whole, right — I verified limits (25/25 both directions), texture formats (round-trip over all 118 header enumerants), features (78/78 symmetric), downlevel flags (26/26), and the counter/HDR/allocator-report ownership protocols as bit-exact. The defects are concentrated in the *header*: values C callers cannot name, initializers that don't initialize, and one feature-gate that doesn't build.

---

## Correctness

### 1. `wgsl` feature does not enable `naga` — build break — [[Cargo.toml:66](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/Cargo.toml:66)](.agent-review/pr-617/workspace/Cargo.toml:66)

`wgsl = ["wgc/wgsl"]`, but [[src/lib.rs:5189](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:5189)](.agent-review/pr-617/workspace/src/lib.rs:5189) now does `use naga::front::wgsl::ImplementedLanguageExtension;` under `#[cfg(feature = "wgsl")]`. `naga` is an optional direct dependency, so enabling `wgsl` without also enabling a feature that pulls `dep:naga` fails.

```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `naga`
    --> src/lib.rs:5189:9
```

**Validated:** `cargo check --no-default-features --features metal,wgsl` fails; adding `dep:naga` to the feature list fixes it (change reverted). Neither CI job (`cargo check`, `cargo check --no-default-features`) exercises this combination, so it will land green.

### 2. `WGPU_NATIVE_LIMITS_INIT` initializes 4 of 25 fields — [[ffi/wgpu.h:1452](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:1452)](.agent-review/pr-617/workspace/ffi/wgpu.h:1452)

The macro names only `chain`, `maxNonSamplerBindings`, `maxBindingArrayElementsPerShaderStage`, `maxBindingArraySamplerElementsPerShaderStage`, `maxMultiviewViewCount`. All 21 new fields — every mesh-shader, ray-tracing, and `maxBindingArrayAccelerationStructureElementsPerShaderStage` limit — get C's zero-fill instead of `WGPU_LIMIT_U32_UNDEFINED`. [[src/conv.rs:769](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:769)](.agent-review/pr-617/workspace/src/conv.rs:769) only skips a field when it equals the sentinel, so **0 is read as an explicit request for 0**.

This does not fail loudly. `Limits::check_limits_with_fail_fn` ([[wgpu-types/src/limits.rs:875](https://github.com/gfx-rs/wgpu)](https://github.com/gfx-rs/wgpu)) computes `invalid_ord = $ordering.reverse()`, so for a max limit it only fails when the request *exceeds* what's allowed. Requesting 0 passes. Device creation succeeds, then wgpu-core validates every BLAS/TLAS/mesh operation against effective limits of 0 — the entire feature set this PR adds is silently unusable for any caller who used the `_INIT` macro as documented.

Separately, the macro is a hard **compile error** under the repo's own example flags:
```
error: missing field 'maxBindingArrayAccelerationStructureElementsPerShaderStage'
       initializer [-Werror,-Wmissing-field-initializers]
       note: expanded from macro 'WGPU_NATIVE_LIMITS_INIT'
```

**Validated:** clang with `-Wall -Wextra -Werror` in both C and C++; a test feeding a zero-filled `WGPUNativeLimits` through `conv::map_required_limits` returns 21 zeros and `check_limits(&Limits::default())` returns `true`; traced the fail-fn ordering in the pinned rev.

### 3. External textures cannot be bound; the BGL path panics — [[src/conv.rs:2093](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2093)](.agent-review/pr-617/workspace/src/conv.rs:2093)

`map_bind_group_entry` gained a `WGPUExternalTextureBindingEntry` path ([[conv.rs:1977](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:1977)](.agent-review/pr-617/workspace/src/conv.rs:1977)), but `map_bind_group_layout_entry`'s chained-struct parameter handles only `WGPUAccelerationStructureBindingLayout`, and the `follow_chain!` at [[src/lib.rs:2170](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:2170)](.agent-review/pr-617/workspace/src/lib.rs:2170) never lists `WGPUSType_ExternalTextureBindingLayout` — which `webgpu.h:2186` documents as *required* to accompany the entry. `wgt::BindingType::ExternalTexture` is never constructed anywhere in the crate.

Trigger: any `WGPUBindGroupLayoutEntry` with `WGPUExternalTextureBindingLayout` chained → `panic!("invalid bind group layout entry for bind group layout descriptor")` at [[conv.rs:2098](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2098)](.agent-review/pr-617/workspace/src/conv.rs:2098), out of a plain `extern "C"` fn (abort).

**Validated:** `grep -rn 'BindingType::ExternalTexture\|ExternalTextureBindingLayout' src/` → zero hits (I ran this independently); agent test constructing the entry panics as predicted. So `wgpuDeviceCreateExternalTexture` produces a handle with no way to use it.

### 4. `WGPUBufferUsage_RayTracingPipelineShaderData` is rejected by the mapper — [[ffi/wgpu.h:971](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:971)](.agent-review/pr-617/workspace/ffi/wgpu.h:971)

The header declares `0x0000000000008000`. That is `BufferUses::…` bit 15 (internal HAL), not a `BufferUsages` bit — `BufferUsages::all() == 0xfff`. `from_u64_bits` ([[conv.rs:2504](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2504)](.agent-review/pr-617/workspace/src/conv.rs:2504)) is `bitflags::from_bits`, which returns `None` on unknown bits, and [[src/lib.rs:2212](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:2212)](.agent-review/pr-617/workspace/src/lib.rs:2212) does `.expect("invalid buffer usage")`.

Also: `BLAS_INPUT` (0x400) and `TLAS_INPUT` (0x800) — the usages you actually need to build acceleration structures — have **no named header constant at all**.

**Validated:** temporary `#[test]` asserting `BufferUsages::all().bits() == 0xfff` and `from_u64_bits::<BufferUsages>(0x8000).is_none()`; reproduced the `expect` panic.

### 5. `wgpuAdapterGetPresentationTimestamp` aborts on the "unsupported" sentinel — [[src/lib.rs:1046](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:1046)](.agent-review/pr-617/workspace/src/lib.rs:1046)

```rust
assert!(ts.0 <= u64::MAX as u128, "presentation timestamp {0} ns overflows u64", ts.0);
```
`PresentationTimestamp::INVALID_TIMESTAMP == u128::MAX` (`wgpu-types/src/surface.rs:1061`) and is returned unconditionally by `wgpu-hal/src/gles/adapter.rs:1387`, `noop/mod.rs:250`, and `vulkan/adapter.rs:3244` under `cfg(not(unix))`. So: any GL adapter on any OS, and Vulkan on Windows, abort the process. Plain `extern "C"`, so panic ⇒ abort.

The header comment at [[ffi/wgpu.h:2277](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:2277)](.agent-review/pr-617/workspace/ffi/wgpu.h:2277) ("timestamps will not exceed 2^64 … in practice") models this as an overflow guard. It isn't — the sentinel is deliberately out of range. Needs a documented `UINT64_MAX` passthrough or a validity flag.

**Validated:** read all four hal impls and the `wgt` constant. On this machine's Metal adapter the real path returns a plausible value, so the assert is only reachable on the sentinel backends — not locally executable.

### 6. `WGPULoadOp_DontCare` and `WGPUTextureAspect_Plane*` don't compile in C++ — [[ffi/wgpu.h:962](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:962)](.agent-review/pr-617/workspace/ffi/wgpu.h:962), [[980-982](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:980)](.agent-review/pr-617/workspace/ffi/wgpu.h:980)

They're `#define`d to bare integers, so assigning one to the enum-typed struct field is a hard error in C++:
```
error: assigning to 'WGPULoadOp' from incompatible type 'int'
```
Every C++ consumer of `WGPULoadOp_DontCare` — the marquee `StrictWebgpuCompliance` feature — is broken. This is @Vipitis's unanswered comment, and it is correct.

**Validated:** reproduced the exact diagnostic with `clang++ -std=c++17`; `static const WGPULoadOp WGPULoadOp_DontCare = (WGPULoadOp)0x00030001;` compiles clean in both C and C++.

### 7. `WGPUSType_SurfaceSourceOhosNativeWindow` is dropped — [[ffi/wgpu.h:21-67](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:21)](.agent-review/pr-617/workspace/ffi/wgpu.h:21)

The new `WGPUNativeSType` block skips `0x0003000C` and `0x00030011`. `0x0003000C` is trunk's `WGPUSType_SurfaceSourceOhosNativeWindow` — added four commits ago in [[#626](https://github.com/gfx-rs/wgpu-native/pull/626)](https://github.com/gfx-rs/wgpu-native/pull/626). Merging this as-is removes OpenHarmony surface support with no CHANGELOG entry. The `0x00030011` hole is unexplained.

**Validated:** confirmed `ffi/wgpu.h:43` in trunk is `WGPUSType_SurfaceSourceOhosNativeWindow = 0x0003000C`, absent from the PR head. This is also why the PR is CONFLICTING.

### 8. `wgpuDevicePoll` gains a 4th parameter, and the parameter is unusable — [[ffi/wgpu.h:2538](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:2538)](.agent-review/pr-617/workspace/ffi/wgpu.h:2538), [[src/lib.rs:5248](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:5248)](.agent-review/pr-617/workspace/src/lib.rs:5248)

`WGPUBool wgpuDevicePoll(WGPUDevice, WGPUBool, WGPUSubmissionIndex const*, uint64_t timeout_ns)`. Silent ABI break — existing compiled callers pass three arguments and the callee reads a garbage register as the timeout. **No CHANGELOG entry.**

And when the timeout does fire, `WaitIdleError::Timeout` goes to `handle_error_fatal` → panic. A C caller cannot observe the timeout it requested except by catching a Rust panic. The `0 == wait forever` semantics and the `submission_index.copied()` plumbing are correct; the error routing is not. `wgpuInstancePollAllDevices` did not get the parameter — asymmetric.

Parameter is also the header's only snake_case identifier besides `ui_view` (finding 13).

### 9. `wgpuSurfaceDiscardTexture` aborts on ordinary validation errors — [[src/lib.rs:4919](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:4919)](.agent-review/pr-617/workspace/src/lib.rs:4919)

`handle_error_fatal` out of a plain `extern "C"` fn. `Surface::discard_inner` returns `SurfaceError::NotConfigured` and `SurfaceError::NothingToPresent`, both `ErrorType::Validation`. Trigger: calling it twice, or on the resize path [[the header itself recommends](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:2698)](.agent-review/pr-617/workspace/ffi/wgpu.h:2698) when no texture was acquired that frame. The adjacent `wgpuSurfacePresent` handles the same error class with `log::warn!` + `WGPUStatus_Error`, and `SurfaceData` already carries an `error_sink` this function ignores.

**Validated:** traced to the wgpu-core variants; not executed (needs a real surface).

### 10. `map_texture_aspect` silently falls back to `All` — [[src/conv.rs:70](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:70)](.agent-review/pr-617/workspace/src/conv.rs:70)

The old `map_enum_with_undefined!` expansion ended in `.expect("Unknown texture aspect")`. The hand-written replacement ends in `_ => None`, and all three call sites — [[conv.rs:946](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:946)](.agent-review/pr-617/workspace/src/conv.rs:946), [[lib.rs:1547](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:1547)](.agent-review/pr-617/workspace/src/lib.rs:1547), [[lib.rs:5032](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:5032)](.agent-review/pr-617/workspace/src/lib.rs:5032) — do `.unwrap_or(wgt::TextureAspect::All)`. A garbage `aspect` in `WGPUTexelCopyTextureInfo` now copies the whole texture instead of erroring.

**Validated:** read the macro expansion at `utils.rs:285`; test asserting `map_texture_aspect(0x00030003) == None`. Same pattern at [[conv.rs:2352](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2352)](.agent-review/pr-617/workspace/src/conv.rs:2352) (`map_surface_color_space` → `Auto` for anything outside 0–7), [[conv.rs:2549](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2549)](.agent-review/pr-617/workspace/src/conv.rs:2549) (update mode → `Build`), [[conv.rs:510](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:510)](.agent-review/pr-617/workspace/src/conv.rs:510)/[[518](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:518)](.agent-review/pr-617/workspace/src/conv.rs:518) (memory hints → `Performance`, experimental features → off). Every sibling conversion in the file either returns `Option`/`Result` or panics.

### 11. Two enum values exist only as magic constants in Rust — [[src/conv.rs:246](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:246)](.agent-review/pr-617/workspace/src/conv.rs:246), [[249](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:249)](.agent-review/pr-617/workspace/src/conv.rs:249)

`WGPU_NATIVE_STORAGE_TEXTURE_ACCESS_ATOMIC = 0x00030001` and `WGPU_NATIVE_TEXTURE_FORMAT_R64_UINT = 0x00030009` have **no header enumerant** — `WGPUNativeTextureFormat` jumps `P010 = 0x00030008` → `Astc4x4Sfloat = 0x0003000A`, and there is no `WGPUNativeStorageTextureAccess` enum at all. So `wgt::StorageTextureAccess::Atomic` is reachable from C only by hardcoding the number, even though the adapter can report `WGPUNativeTextureFormatFeatureFlags_StorageAtomic` ([[ffi/wgpu.h:2373](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:2373)](.agent-review/pr-617/workspace/ffi/wgpu.h:2373)).

Worse, `0x00030009` is **emitted** outward: `to_native_texture_format` lost its `None` arms (finding 20), so `wgpuSurfaceGetCapabilities` and `wgpuAdapterGetTextureFormatCapabilities` can hand a caller a value with no symbol to compare against.

**Validated:** `grep 'R64\|StorageTextureAccess' ffi/wgpu.h` → only a feature doc comment, zero enumerants (ran independently); round-trip test on `0x00030009 ↔ TextureFormat::R64Uint`.

### 12. `WGPUAdapterInfoExtras` is ignored unless it is first in the chain — [[src/lib.rs:861](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:861)](.agent-review/pr-617/workspace/src/lib.rs:861), [[898](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:898)](.agent-review/pr-617/workspace/src/lib.rs:898), [[6433](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:6433)](.agent-review/pr-617/workspace/src/lib.rs:6433)

All three sites test `(*info.nextInChain).sType` on the head only. Chain the extras second and `wgpuAdapterGetInfo` returns `Success` with `transientSavesMemory`/`devicePciBusId` untouched, and `wgpuAdapterInfoFreeMembers` won't free them. Inconsistent with `wgpuSurfaceGetCapabilities`/`…FreeMembers` (lib.rs:4808, 4995), which do walk, and with `follow_chain!` used everywhere else. `write_limits_struct` ([[conv.rs:613](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:613)](.agent-review/pr-617/workspace/src/conv.rs:613)) has the same head-only bug — pre-existing, but this PR added 21 fields behind it.

**Validated:** agent test with sentinel values in a second-position extras — survives untouched.

### 13. `WGPUSurfaceSourceUIView` is not platform-gated and is UB on NULL — [[src/conv.rs:2313](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2313)](.agent-review/pr-617/workspace/src/conv.rs:2313)

The metal branch above it is `#[cfg(all(any(target_os = "ios", target_os = "macos"), feature = "metal"))]`; the UIView branch has no cfg, so on Linux/Windows a chained `WGPUSurfaceSourceUIView` builds a `RawWindowHandle::UiKit` and hands it to wgpu instead of erroring. `NonNull::new_unchecked(uiview.ui_view)` is immediate UB for NULL, where `map_native_display_handle` in the same file (conv.rs:463/472/481) uses `NonNull::new(..).expect(..)`.

The field is `void *ui_view` ([[ffi/wgpu.h:1812](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:1812)](.agent-review/pr-617/workspace/ffi/wgpu.h:1812)) — the only snake_case struct field in either header, and new in this PR. Rename to `uiView` before it ships; it's free now and API-breaking later.

### 14. Passthrough shader descriptors pass `{NULL, 0}` string views

Agent-verified live on Metal. `wgpuDeviceCreateShaderModulePassthrough` builds `StringView`s from empty labels as `{NULL, 0}` rather than the repo's convention, which downstream code distinguishes from a genuinely empty label.

**Validated by subagent** with a live Metal device; I did not re-execute this one.

### 15. `wgpuExternalTextureAddRef` / `Release` are declared twice — [[ffi/wgpu.h:2623-2624](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:2623)](.agent-review/pr-617/workspace/ffi/wgpu.h:2623)

`webgpu.h:6384-6385` already declares both **with** `WGPU_EXPORT WGPU_FUNCTION_ATTRIBUTE`. `ffi/wgpu.h` redeclares them without. These are the only two duplicates across the two headers (I diffed the full symbol sets). On Windows shared-library builds `WGPU_EXPORT` expands to `__declspec(dllimport)`, and a redeclaration without it triggers `-Winconsistent-dllimport` / lets MSVC bind the non-imported form.

**Validated:** duplicate confirmed by symbol-set diff. The Windows diagnostic itself is **not locally reproduced** — no Windows SDK on this machine. Note that `ffi/wgpu.h` omitting `WGPU_EXPORT` entirely is pre-existing in trunk, so only the duplication is new.

### 16. `wgpuDeviceGetInternalCounters` always returns all zeros — [[src/lib.rs:6051](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:6051)](.agent-review/pr-617/workspace/src/lib.rs:6051)

`InternalCounter::read()` is `#[cfg(not(feature = "counters"))] -> 0`, and `Cargo.toml` has no `counters` feature and never enables `wgc/counters` — so this is unreachable-by-configuration even under `--all-features`. The header ([[ffi/wgpu.h:2655](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:2655)](.agent-review/pr-617/workspace/ffi/wgpu.h:2655)) advertises it as a debugging facility with no caveat.

**Validated:** read the cfg in the pinned rev and grepped `Cargo.toml` (both independently); agent created a real buffer on a live Metal device and observed `hal.buffers == 0`.

This is the substance behind @Vipitis's "`WGPUInternalCounters` looks like an empty placeholder" — but note the reviewer's specific premise was wrong: `wgt::CoreCounters` is genuinely empty upstream, so dropping the `core` field is fine. The problem is the feature gate, not the struct.

### 17. `extern "C-unwind"` applied to 14 of 255 exported functions

Trunk has zero. This PR adds `C-unwind` to exactly 14 (lib.rs:1136, 1162, 2458, 3323, 3331, 3383, 3534, 3663, 3885, 4685, 4841, 5155, 5217, 5248) and leaves 241 plain. There's no stated rule for which get it, and the selection doesn't track "can panic" — findings 3, 5, 9 are all abort-on-panic paths that didn't get it, while `wgpuInstancePollAllDevices` did. Either mark all of them or none; a 14/241 split is a coin-flip on whether a given panic unwinds into C or aborts.

### 18. `WGPUBindGroupEntryExtras` with `size == 0` now panics — [[src/conv.rs:1944](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:1944)](.agent-review/pr-617/workspace/src/conv.rs:1944)

Was `if entry.size == 0 { None } else { Some(entry.size) }`; now `0 => panic!("invalid size")`. Aligning the array path with the single-buffer path (conv.rs:1890) and with `WGPU_BIND_GROUP_ENTRY_INIT`'s `size = WGPU_WHOLE_SIZE` is the right call, but existing C code that passed `0` to mean "whole buffer" now aborts. **No CHANGELOG entry.**

### 19. `wgpuInstancePollAllDevices` uses Rust `bool` against a `WGPUBool` header — [[src/lib.rs:3534](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:3534)](.agent-review/pr-617/workspace/src/lib.rs:3534)

`WGPUBool` is `uint32_t`; Rust `bool` params/returns are `noundef zeroext i1`. A caller passing `wait = flags & MASK` is UB. Pre-existing on `wgpuDevicePoll` and `wgpuDeviceStartGraphicsDebuggerCapture` — this adds a third rather than fixing the pattern. Every other exported fn uses `native::WGPUBool`.

### 20. `to_native_texture_format` can no longer return `None`, and its caller still filters — [[src/conv.rs:1313](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:1313)](.agent-review/pr-617/workspace/src/conv.rs:1313)

The PR deleted the `Astc{Hdr} => None` and `R64Uint => None` arms; the match now has **zero** `_ =>` arms and covers every `TextureFormat`. [[src/lib.rs:4757](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:4757)](.agent-review/pr-617/workspace/src/lib.rs:4757) still reads *"some texture formats are not in webgpu.h and `conv::to_native_texture_format` returns None for them. so, filter them out"* and calls `filter_map`, now a no-op. The `Option` return is dead; combined with finding 11, the filter no longer shields C callers from undeclared values.

**Validated:** `awk` count of `_ =>` over lines 1311–1440 → 0 (ran independently); test asserting `is_some()` for R64Uint, ASTC-Hdr, NV12, P010.

### 21. Header hygiene

- The 14 new ASTC-Sfloat enumerants have no `_Force32` member — [[ffi/wgpu.h:2259-2275](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:2259)](.agent-review/pr-617/workspace/ffi/wgpu.h:2259). Every other enum in both headers has one.
- Broken doxygen refs: `WGPUStringView_null` ([[1522](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:1522)](.agent-review/pr-617/workspace/ffi/wgpu.h:1522), 1527, 1529), `WGPUNativeFeature_MultiView` ([[1964](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:1964)](.agent-review/pr-617/workspace/ffi/wgpu.h:1964) — the enumerant is `Multiview`), `wgpuRayTracingPipeline*` ([[836](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:836)](.agent-review/pr-617/workspace/ffi/wgpu.h:836) — no such functions exist).
- `extras.devicePciBusId` when unavailable is `{data = 0x1, length = 0}`, not `{NULL, 0}` — the dangling pointer from `String::from("").into_boxed_str()`. Header says "Empty when not available"; a caller testing `data != NULL` mis-detects. Freeing it is benign.

---

## Assertions

**CHANGELOG:** *"Wgpu-native has reached feature-parity with upstream wgpu. … Now tested against wgpu's test suite."* — [[CHANGELOG.md:15](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/CHANGELOG.md:15)](.agent-review/pr-617/workspace/CHANGELOG.md:15)

Three problems, each independently disqualifying:
- The PR title is "**part 1**", and `src/unimplemented.rs` still has 6 live `unimplemented!()` (`wgpuGetProcAddress`, both `Create*PipelineAsync`, `wgpuDeviceGetLostFuture`, `wgpuInstanceWaitAny`, `wgpuTextureGetTextureBindingViewDimension`). Not parity.
- "Now tested against wgpu's test suite" is contradicted by the PR description's own *"It was tested but that won't be part of this PR and will come later"* — and the diff adds **zero** test files (verified against the file list in `pr-diff.diff`).
- Both new entries credit **#594**, not #617.

**CHANGELOG:** *"Updated all wgpu crates to v29"* — [[CHANGELOG.md:29](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/CHANGELOG.md:29)](.agent-review/pr-617/workspace/CHANGELOG.md:29). `Cargo.toml` now pins a rev four weeks past v30.0.0. Stale line, no v30 entry added.

**Missing CHANGELOG entries** for three breaking changes: `wgpuDevicePoll`'s 4th parameter (finding 8), `WGPUBindGroupEntryExtras.size == 0` (finding 18), and the dropped OpenHarmony SType (finding 7).

**Header:** *"timestamps will not exceed 2^64 nanoseconds in practice"* — [[ffi/wgpu.h:2277](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/ffi/wgpu.h:2277)](.agent-review/pr-617/workspace/ffi/wgpu.h:2277). The value that *does* exceed it is a deliberate sentinel, returned by three backends. See finding 5.

**Stale comment:** *"These helpers centralise the cross-platform casts"* — [[src/lib.rs:1199](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:1199)](.agent-review/pr-617/workspace/src/lib.rs:1199), sitting above a `#[allow(clippy::unnecessary_cast)]` on `wgpuBufferMapAsync`. There are no helpers; there is one cast.

**Stale comment:** the `filter_map` rationale at [[src/lib.rs:4757](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:4757)](.agent-review/pr-617/workspace/src/lib.rs:4757) — finding 20.

**Accurate claim, verified:** *"wgpu-core does not expose a set-label API for these resource types, so these are permanent no-ops"* — [[src/lib.rs:6508](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:6508)](.agent-review/pr-617/workspace/src/lib.rs:6508). Checked against the pinned rev; all ~18 are honest.

**Missing assertion:** [[conv.rs:2106](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2106)](.agent-review/pr-617/workspace/src/conv.rs:2106) does `extras.and_then(…count).or_else(|| …bindingArraySize)`, silently preferring the extras field. Both mean binding-array size; two conflicting non-zero values produce no diagnostic, and `count = 0` falls through. Needs a comment or an assert.

**Reachable panics from non-panicking C APIs**, beyond the ones above: [[lib.rs:2212](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:2212)](.agent-review/pr-617/workspace/src/lib.rs:2212) `expect("invalid buffer usage")`, [[conv.rs:2098](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2098)](.agent-review/pr-617/workspace/src/conv.rs:2098) invalid-BGL-entry, [[conv.rs:2335](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2335)](.agent-review/pr-617/workspace/src/conv.rs:2335) `panic!("Error: Unsupported Surface")` — the last one fires for `WGPUSurfaceSourceDrm` on any build without the non-default `drm` feature, i.e. every shipped `cdylib`.

---

## LLM-isms

- **`map_state_to_u32` exists only to carry a lint suppression** — [[conv.rs:2628](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:2628)](.agent-review/pr-617/workspace/src/conv.rs:2628). `native::WGPUBufferMapState` is `c_uint` in the generated bindings, so `s as u32` really is a no-op; removing `#[allow(clippy::unnecessary_cast)]` makes clippy fire. If it's there for MSVC's `c_int`, that needs the one-line rationale every other suppression in the tree has (cf. `utils.rs:323`). *Cosmetic, but the suppression hides why the wrapper exists.*
- **Verbatim copy-paste pair** — `wgpuAdapterGetInfo` ([[lib.rs:839](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:839)](.agent-review/pr-617/workspace/src/lib.rs:839)) and `wgpuDeviceGetAdapterInfo` ([[lib.rs:6412](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:6412)](.agent-review/pr-617/workspace/src/lib.rs:6412)) are character-identical. No missed substitution, but they share finding 12's bug and will drift.
- **Duplicated BLAS index logic** — [[lib.rs:5889-5902](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:5889)](.agent-review/pr-617/workspace/src/lib.rs:5889) and [[lib.rs:6251-6263](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:6251)](.agent-review/pr-617/workspace/src/lib.rs:6251). Diffed; currently identical. *Points at nothing yet; will.*
- **Reimplementation of existing helpers** — [[lib.rs:1008](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:1008)](.agent-review/pr-617/workspace/src/lib.rs:1008) re-derives `conv::to_native_texture_usage_flags` inline (bit layouts coincide today; latent). [[lib.rs:1466-1479](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:1466)](.agent-review/pr-617/workspace/src/lib.rs:1466) and [[lib.rs:2468-2481](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:2468)](.agent-review/pr-617/workspace/src/lib.rs:2468) hand-roll `nextInChain` loops for `multiviewMask` where `follow_chain!` is used for the same job at lib.rs:2528/2722 — and skip `follow_chain!`'s alignment assertion before the reinterpret-cast. Same at lib.rs:4811 and 4998.
- **Misleading comment** — [[conv.rs:245](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:245)](.agent-review/pr-617/workspace/src/conv.rs:245) *"native extension, after standard StorageTextureAccess range"*: the standard range is 0–4, and `0x00030001` is the wgpu-native namespace, not "after". An unrelated texture-format constant is parked in the storage-texture-access section on the next line.
- **Prose tells in the CHANGELOG** — "reached feature-parity", "Now tested against wgpu's test suite". Both false; see Assertions.
- **Enum-shaped values written as `#define`** — findings 6 and 11. This is the classic tell of generating header entries by string-appending rather than editing the enum body, and it produced two real defects.
- **`deny.toml` skip without a version pin** — `{ crate = "linux-raw-sys", reason = "drm crate depends on multiple versions" }`. Every sibling entry pins `@version`; this one will silently absorb future duplicates.

**Not slop** (checked, ruled out): the `WebGpuError for ShaderParseError` impl is required by `handle_error`'s bound; `DisplayGamut`'s `Some(_)` arm is required because the enum is `#[non_exhaustive]`; non-const `nextInChain` matches upstream webgpu.h (contra the review thread); `enable wgpu_binding_array` / `@interpolate(flat)` in the example WGSL are both genuinely required now.

**Genuine fixes worth calling out:** the `handle_error` rework to `webgpu_error_type()` ([[lib.rs:674](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:674)](.agent-review/pr-617/workspace/src/lib.rs:674)), the `Option<VertexBufferLayout>` mapping that removes the unused-vertex-buffer panic ([[lib.rs:2564](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:2564)](.agent-review/pr-617/workspace/src/lib.rs:2564)), `COPY_DST` no longer emitting `WGPUTextureUsage_CopySrc`, and `budgetForDeviceCreation` → `budgetForDeviceLoss` ([[conv.rs:418](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/conv.rs:418)](.agent-review/pr-617/workspace/src/conv.rs:418)).

---

## Tests

Zero tests in a +4207-line FFI change. Concretely, findings 1, 2, 4, 6, and 11 would each be caught by a test that costs under ten lines:

- **A C and C++ TU that instantiates every `_INIT` macro and asserts each field equals its documented default.** Catches finding 2 outright, and finding 6 by simply existing as C++ — there is currently no C++ compile check at all, which is why `WGPULoadOp_DontCare` shipped unusable.
- **A round-trip test over every declared header enum value.** Catches finding 11 (values in `conv.rs` with no header enumerant, and vice versa) and finding 20.
- **A `from_u64_bits` test over every named `WGPUBufferUsage_*` / `WGPUTextureUsage_*` constant.** Catches finding 4.
- **A CI job for each individually-selectable feature**, not just all-on and all-off. Catches finding 1.
- **Nothing pins the ABI coupling** at [[lib.rs:1009](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:1009)](.agent-review/pr-617/workspace/src/lib.rs:1009), where `wgt::TextureFormatFeatureFlags::bits()` is passed straight through as `WGPUNativeTextureFormatFeatureFlags`. All 11 bits align today by hand-check; a `const _: () = assert!(...)` per bit would keep it that way.

---

## Unverified

- **Finding 15's Windows diagnostic** — no Windows SDK here (`clang --target=x86_64-pc-windows-msvc` can't find `math.h`). *Settled by:* one clang-cl or MSVC compile of both headers with `-DWGPU_SHARED_LIBRARY`.
- **Findings 5, 9, 13** are macOS-unreachable. *Settled by:* a Linux run (GL adapter for 5, `WGPUSurfaceSourceUIView` for 13) and a C example that configures a surface and calls `wgpuSurfaceDiscardTexture` twice (9).
- **Finding 8's timeout panic** is proven by code path, not execution — an idle device returns immediately. *Settled by:* submitting a multi-second workload and calling `wgpuDevicePoll(device, true, NULL, 1)`.
- **`wgpuGetWgslLanguageFeatures` vs `wgpuInstanceGetWGSLLanguageFeatures` disagree by construction** — the former returns `0xF` including `ImmediateAddressSpace`; the latter enumerates 3 because there is no matching `WGPUWGSLLanguageFeatureName` (none found in 0x1..0xA). *Settled by:* confirming upstream has no enumerant, then documenting the asymmetry or adding a native one.
- **`wgpuSurfaceGetDisplayHdrInfo` was never executed.** wgpu-core documents that the Metal backend returns `default()` when queried off the main thread; the wrapper's doc comment ([[lib.rs:4949](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:4949)](.agent-review/pr-617/workspace/src/lib.rs:4949)) doesn't relay that. *Settled by:* a surface-backed example on Windows (DXGI is the only backend that populates `luminance`/`chromaticity`) and macOS off-thread.
- **`map_features` drops unmapped names silently**, so a C caller requesting `WGPUFeatureName_Subgroups`, `CoreFeaturesAndLimits`, `TextureComponentSwizzle`, or `TextureFormatsTier1/2` gets a no-op instead of a failed device creation. Pre-existing `_ => None`. *Settled by:* deciding whether unknown requested features should be an error.
- **`wgt::VertexFormat::Float64{,x2,x3,x4}` have no webgpu.h value**, so `WGPUNativeFeature_VertexAttribute64bit` is advertisable but unusable. Pre-existing; wants `WGPUNativeVertexFormat` values.
- **`Cargo.lock` reverts 9 unrelated dependency bumps** from trunk's renovate PRs #613/#620. Almost certainly a rebase artifact rather than intent, but it is in the diff. *Settled by:* the author confirming and re-running `cargo update` for those crates.
- **`has_surface_presented` is a single `Arc<AtomicBool>`** shared by the surface and every texture it hands out (lib.rs:434-448 vs 4861), so releasing a stale texture after a newer `wgpuSurfaceGetCurrentTexture` discards the *current* texture. Pre-existing; `wgpuSurfaceDiscardTexture` extends the mechanism rather than fixing it. *Settled by:* a two-frame C repro holding frame N's texture past frame N+1's acquire.
- **`wgpuAdapterInfoFreeMembers` is asymmetrically idempotent** — [[lib.rs:892](https://claude.ai/epitaxy/.agent-review/pr-617/workspace/src/lib.rs:892)](.agent-review/pr-617/workspace/src/lib.rs:892) nulls `extras.devicePciBusId` but cannot null the by-value `vendor`/`device`/`description`, so a double call is a no-op for extras and a double free for the base strings. Pre-existing for the base fields; the new extras path establishes the opposite convention inside the same function.

**Verified clean**, so you can skip re-checking: `cargo clippy --all-features`, `cargo doc --all-features --no-deps`, and `python3 checks.py` all pass. All 262 declared symbols are exported, none extra. All 29 removed `unimplemented!()` stubs have real definitions. Limits, texture formats, features, downlevel flags, `ShaderRuntimeChecks`, `AccelerationStructureFlags`, `SurfaceColorSpace(s)`, `CooperativeScalarType`, `ShaderModel`, `ExternalTextureFormat`, `VertexFormat`, `InstanceFlags`, `HalCounters`, `CooperativeMatrixProperties` — all mapped exhaustively and bit-exact in both directions. The cooperative-matrix `fill = capacity.min(count)` bound is correct (canary-tested). The allocator-report ownership protocol is sound (`into_boxed_slice` → `Box::into_raw` matched by `Box::from_raw(slice_from_raw_parts_mut)`, so no `Vec` capacity hazard). `StrictWebgpuCompliance`'s two header claims both hold against wgpu-core.

</details>

inner-daemons:
Closing in favor of #630 and #631. Much thanks to @matthargett. I may reopen this myself if his doesn't land for a while but most likely he will take this project over from me.

PR #619, opened 2026-07-12
Expose DeviceDescriptor::memory_hints via WGPUDeviceExtras

RubyBit:
## Checklist

- [x] `cargo doc` reports no issues (unrelated to my changes)

## Description

wgpu's DeviceDescriptor has had memory_hints since MemoryHints was introduced, but map_device_descriptor hardcoded Default::default() (Performance) to be changed in the future (so I changed it): Performance hints make gpu-allocator commit 128 MiB device + 64 MiB host memory blocks for the few tiny internal allocations wgpu makes at device open, so creating a bare device commits ~200 MiB of VRAM (measured on NVIDIA RTX 4080 Laptop/Vulkan; D3D12 is similar), and all committed memory rounds up in 256 MiB steps. Applications with small or mostly-static workloads have had no way to opt out through the C API (such as my tensor library which can load pretty small models on the gpu through wgpu and there is an unnecessary, around 200 MiB overhead). 

Add a WGPUMemoryHints enum (Undefined/Performance/MemoryUsage/Manual) and memoryHints + suballocatedDeviceMemoryBlockSize{Start,End} fields to WGPUDeviceExtras, mapped onto wgt::MemoryHints. Zero-initialized extras keep the previous behavior (Undefined -> Performance).

Measured bare-device committed VRAM (RTX 4080 Laptop, Vulkan):
  Performance (old default): 201 MiB
  MemoryUsage:                21 MiB
  Manual { 4 MiB..256 MiB }:  17 MiB

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The format is loosely based on [Keep a Changelog](https://keepachangelog.com/en/
 
 ## Unreleased
 
+### Added
+- `WGPUDeviceExtras` now exposes `DeviceDescriptor::memory_hints` via `memoryHints` (`WGPUMemoryHints`: `Performance`/`MemoryUsage`/`Manual` with a block-size range); zero-initialized keeps the previous behavior. By @RubyBit in [#619](https://github.com/gfx-rs/wgpu-native/pull/619).
+
 ### Fixed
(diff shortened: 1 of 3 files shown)


inner-daemons:
Added by #594 so unless you need this instantly I would prefer to wait.

RubyBit:
I see, but when can we expect the PR merged then? v30 also released a few weeks ago.

inner-daemons:
@RubyBit I would expect within the next month at most for it to be merged, but I have hopes it could happen sooner.

cwfitzgerald (approved):
Thanks!

PR #632, opened 2026-09-16
WIP: Update wgpu.yml

QuantumSegfault:
## Checklist

Not addressed yet. TODO.

- [ ] `cargo clippy` reports no issues
- [ ] Human-readable change descriptions added to CHANGELOG.md under the "Unreleased" heading.
  - [ ] Add credit to yourself for each change: `Added new functionality. @githubname`

## Description

This is a WIP effort to bring `wgpu.yml` up to speed with changes in `wgpu.h`, and to have the header generated by the `webgpu-headers` bindings generator (and thus be work with other compatible generators for other languages).

The goal for this PR is to have zero ABI or API changes. However this is required a small patch to `webgpu-headers`, and some hacks/uncleanliness in the YAML to accomplish.

This also requires a handful of manual touchups (labeled with `FIXME` in the YAML `doc`). See 23632c79c60e65f9a7f832c8d29c336ad90ac991


QuantumSegfault:
Before I sink any more time into this, is there community interest in this? I assume there's at least some, given #606 (that we have this YAML at all).

Also, what's the policy on breaking changes. Will there be opportunities to clean up the interface in the near future?

QuantumSegfault:
@cwfitzgerald @Vipitis @jalexcole 

?

Vipitis:
Hey, thanks for looking into this.

There is definitely interest as we have been discussing some codegen improvements for `wgpu-py` and it would be easier to go from a yaml or json instead of parsing the header files. 

If there is time, perhaps it can be discussed in [todays maintainer meeting].(https://docs.google.com/document/u/0/d/1Z3qjy3m7eAYaTsh2n-iKxLV4Hjc6wZxgukzdQOgVH1c) There is a large PR coming already that adds a most of the wgpu native only features. I am not sure how soon that will merge... But it might be worth to wait as it includes quite a few additions to `wgpu.h`. So breaking changes are happening either way - would be nice if they are documented in the changelog.

If you need tweaks for the webgpu-headers script, perhaps open an issue or PR there. But as far as I know they do not accept any breaking changes to `webgpu.h` anymore.

jalexcole:
@QuantumSegfault I made this yaml to work on java bindings to webgpu. This allows for new enums and classes to be generated in the same way that I generate from webgpu.yaml.

QuantumSegfault:
Okay. Good to know!

> There is definitely interest as we have been discussing some codegen improvements for wgpu-py and it would be easier to go from a yaml or json instead of parsing the header files.

Yeah. I'm tinkering around with a new set of more (automated) bindings for D, based on the official codegen. But `wgpu` here is a missing piece of the puzzle.

I personally use the JSON to parse.

> But it might be worth to wait as it includes quite a few additions to wgpu.h. So breaking changes are happening either way - would be nice if they are documented in the changelog.

Oh...hmmm...

What kind of breaking changes are we allowing? API only? Or ABI as well? Because it would be nice to undo all the hacks. 

Ultimately that would change the name of most symbols (using a `Wgpu` namespace consistently; so stuff like `wgpuDeviceWgpuPoll`, `WGPUFeatureName_WgpuXXXX`, `WGPUWgpuPipelineStatisticName`, `wgpuWgpuGenerateReport`). Or could still be `native` instead. Though if that's too ugly, the ability to use `namespace: webgpu` to remove prefixes is vanilla. See https://github.com/webgpu-native/webgpu-headers/blob/main/tests/extensions/extension.yml. This would be both ABI and API breaking (ABI only cause of the name of functions changing).

It would also enforce the use of our `0x0003_` enum prefix allocation consistently. ALL enums would have our namespace prefix, even ones that are custom (I had to modify `webgpu-headers` to not do this). This would be ABI breaking only.

There are other minor tweaks to names that would be API only, replacing `GLSL` with `Glsl` (so the YAML doesn't need to be spelled out `_g_l_s_l`. 

The log callback would have different API and ABI

There's also a few spots where the pointer and count of an "array" are swapped compared to `webgpu` standard, so that would also be breakage. 

> If you need tweaks for the webgpu-headers script, perhaps open an issue or PR there. But as far as I know they do not accept any breaking changes to webgpu.h anymore.

If we can adapt to their system, then there's only a small non-breaking change that needs to be made, adding support for `uint8_t` (see `instance_extras`). But we might not even need that if we just use `void` (for a `void*` for `budget_for_device_creation` and `budget_for_device_loss`; same as the current YAML actually!).

I mainly modified the script to allow our non-standard header to be represented a bit easier.

------

> I made this yaml to work on java bindings to webgpu. This allows for new enums and classes to be generated in the same way that I generate from webgpu.yaml.

Yup. That's my hope too. For D.

QuantumSegfault:
Alright. I've finished this up.

There's just one, hopefully minor, breakage, with extension enums no longer having their own dedicated enum typedef. If that's a deal-breaker, I can manually patch that part for now as well until breakage is allowed.

PR #635, opened 2026-09-21
add exportable buffer

clewis7:
## Checklist

  - [x] Add credit to yourself and the PR for each change: `Added new functionality. By @githubname in [#XYZ](https://github.com/gfx-rs/wgpu-native/pull/XYZ).`

## Description

Hi @almarklein :D 

As you know, @kushalkolar and I have been working on some GPU-resident compute and viz. He on the JPEG encoding and I more so on model interpretability. I think it would be easier for both of us if we could have this exportable buffer stuff integrated upstream. 

Making a PR from my fork here. Happy to make any changes :) 

Thanks!


Vipitis:
Hey Caitlin,

It's exciting to see the usecase here: https://clewis7.github.io/research/pytorch-wgpu/ and I am surprised it works. However I worry it might not fit as a single backend, single platform feature here. The wgpu-core interoperability is a larger discussion: https://github.com/gfx-rs/wgpu/issues/4067

I wonder if the same can be achieved with external texture? Not familiar enough with pytorch internals or the cuda texture API, but external texture is part of the webgpu spec. (And it's also included in the v30 PR). 
Can't comment in the rust implementation.

kushalkolar:
Also curious how this is different from this :thinking: https://github.com/gfx-rs/wgpu/discussions/7988

almarklein:
> However I worry it might not fit as a single backend, single platform feature here. 

I'm worried a bit about that too, but the proper implementation will probably take a while to land. This'd open up some use-cases in that direction - even if it's just Linux. But we should probably mark it experimental and/or add "linux" to the function name to make the scope clear.

Also curious whether external textures can be used instead. I'll need to do more digging to tell what makes the most sense from an API perspective ...

